PR #9, opened 2022-08-17
Feature/tm sconverer

cmmarshall:
TMS reco/matching merge


src/reco/NDLArTMSMatchRecoFiller.h
```diff
@@ -25,6 +28,27 @@ namespace cafmaker
...
+  
+   TF1* DrawLines(float x1, float z1, float x2, float z2, float startpoint, float endpoint)
+  {
+    TF1 *fun = new TF1("fun","(x - [0])*([3] - [2])/([1] - [0]) + [2]", startpoint, endpoint);
+    fun->SetParameter(0,z1);
+    fun->SetParameter(1,z2);
+    fun->SetParameter(2,x1);
+    fun->SetParameter(3,x2);
+
+    return fun;
```
cjbackhouse, line 40: This seems like a lot of overkill just to calculate this ratio and then evaluate it at a single value of x. Can we just inline this equation as real code in the one place it's used?

src/reco/NDLArTMSMatchRecoFiller.h
```diff
@@ -25,6 +28,27 @@ namespace cafmaker
...
+  }
+
+  float x1_lar, x2_lar, y1_lar, y2_lar, z1_lar, z2_lar;
+  float x1_tms, x2_tms, y1_tms, y2_tms, z1_tms, z2_tms;
+  float x_LAr, x_TMS;
+  float slope_LAr, slope_TMS;
+  float slope_TMS_LAr;
+  float ang;
+  float residual;
+  float Costheta;
```
cjbackhouse, line 50: Are these all necessary? Should they not be local variables in the FillReco function?

src/reco/TMSRecoBranchFiller.h
```diff
@@ -6,24 +6,55 @@
...
+      int _nHitsInTrack[10];
+      float _TrackLength[10];
+      float _TotalTrackEnergy[10];
+      float _Occupancy[10];
+      float _DirectionX_Downstream[10];
+      float _DirectionZ_Downstream[10];
+      float _DirectionX_Upstream[10];
+      float _DirectionZ_Upstream[10];
+      // [10][200][2] needs to match TMS reco output (check file if in doubt)
+      float _TrackHitPos[10][200][2];
```
cjbackhouse, line 57: Personally, I think making these fields that SetBranchAddress is reading into global variables in the .cxx makes more sense than making them class members.

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -3,21 +3,78 @@
...
   // ---------------------------------------------------------------------------
 
+  // here we copy all the TMS reco into the SRTMS branch of the StandardRecord object.
   void TMSRecoBranchFiller::_FillRecoBranches(std::size_t evtIdx,
-                                              caf::StandardRecord &sr,
-                                              const cafmaker::dumpTree &dt,
-                                              const cafmaker::Params &par) const
+      caf::StandardRecord &sr,
+      const cafmaker::dumpTree &dt,
+      const cafmaker::Params &par) const
```
cjbackhouse, line 38: Why did this get reindented?

cjbackhouse (commented):
Looks basically sane to me. We decided to sequence this after the similar SAND merge since they will likely slightly conflict.

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -3,21 +3,78 @@
...
+      TMSRecoTree->SetBranchAddress("nHitsInTrack", _nHitsInTrack);
+      TMSRecoTree->SetBranchAddress("TrackLength", _TrackLength);
+      TMSRecoTree->SetBranchAddress("TotalTrackEnergy", _TotalTrackEnergy);
+      TMSRecoTree->SetBranchAddress("Occupancy", _Occupancy);
+
+      TMSRecoTree->SetBranchAddress("DirectionX_Upstream", _DirectionX_Upstream);
+      TMSRecoTree->SetBranchAddress("DirectionZ_Upstream", _DirectionZ_Upstream);
+
+      TMSRecoTree->SetBranchAddress("DirectionX_Downstream", _DirectionX_Downstream);
+      TMSRecoTree->SetBranchAddress("DirectionZ_Downstream", _DirectionZ_Downstream);
```
cjbackhouse, line 23: Could check the return values here to ensure we don't accidentally miss a format change in the upstream file.

src/reco/NDLArProductFiller.cxx
```diff
@@ -12,7 +12,7 @@ namespace cafmaker
   template <>
   void NDLArProductFiller<caf::SRTrack>::FillSR(caf::StandardRecord &sr, std::size_t evtIdx) const
   {
-    std::cout << "Filling tracks for event " << evtIdx << std::endl;
+    //std::cout << "Filling tracks for event " << evtIdx << std::endl;
```
chenel, line 15: Commenting this (and all the other similar lines below) out is fine, but maybe we should also explore some kind of logging facility?  (Not for this PR, but just asking to gauge the level of interest.)

src/reco/TMSRecoBranchFiller.h
```diff
@@ -6,24 +6,55 @@
...
 
     private:
       void _FillRecoBranches(std::size_t evtIdx,
                              caf::StandardRecord &sr,
                              const dumpTree &dt,
                              const cafmaker::Params &par) const override;
 
-      TFile fTMSRecoFile;
+      TFile *fTMSRecoFile;
+      TTree *TMSRecoTree;
```
chenel, line 43: personally I like `std::unique_ptr<>` rather than having to manually `delete` pointers, but I won't die on that hill

src/reco/NDLArTMSMatchRecoFiller.h
chenel: especially because TF1s are notoriously slow...

chenel (approved):
generally ok---a few suggestions and questions but once Chris's comments are addressed I think we're ready to merge

cjbackhouse:
I pushed the fixes that make this work with the newly-merged SAND branch

cjbackhouse:
Is someone working on the most major of the comments above?

src/makeCAF.C
```diff
@@ -114,30 +114,29 @@ std::vector<std::unique_ptr<cafmaker::IRecoBranchFiller>> getRecoFillers(const c
...
   std::vector<std::unique_ptr<cafmaker::IRecoBranchFiller>> recoFillers;
 
+  // use the run+subrun numbers to seed the random number generator if seed is not explicitly provided
+  // (yes, leak this pointer.  there's only one and sending it back to the main() is annoying)
+  auto rando = new TRandom3(par().cafmaker().seed() >= 0 ?
+                            par().cafmaker().seed() :
+                            par().runInfo().run() * 1000 + par().runInfo().subrun());
+
+  recoFillers.emplace_back(std::make_unique<cafmaker::ParameterizedRecoBranchFiller>(rando));
+
```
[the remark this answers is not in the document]
cjbackhouse: In the brave new world parameterized reco just doesn't exist anymore, so this is moot.

clarenc3:
@cjbackhouse I plan to address them either today or over the weekend. I agree with all the above comments for what it's worth. 

Faiza is updating the track matcher so I'll wait for her and let you know

cfg/ndcafmakerjob.fcl
```diff
@@ -2,7 +2,15 @@
...
+#nd_cafmaker.CAFMakerSettings.InputDumpFile: "/dune/data/users/jwolcott/nd/nd-lar-reco/cafmaker_input/neutrino.LArTMSProductionJun23withLArCV.99.dump.root"
+#nd_cafmaker.CAFMakerSettings.InputGHEPFile: "/dune/data/users/jwolcott/nd/nd-lar-reco/cafmaker_input/neutrino.LArTMSProductionJun23withLArCV.99.ghep.root"
+#nd_cafmaker.CAFMakerSettings.OutputFile: "/dune/data/users/jwolcott/scratch/neutrino.LArTMSProductionJun23withLArCV.99.caf.root"
+#nd_cafmaker.CAFMakerSettings.NDLArRecoFile: "/dune/data/users/jwolcott/nd/nd-lar-reco/cafmaker_input/neutrino.LArTMSProductionJun23withLArCV.99.summary.h5"
+
+# CWret test setup
+nd_cafmaker.CAFMakerSettings.InputDumpFile: "/dune/app/users/cwret/ND_CAFMaker/death.root"
+nd_cafmaker.CAFMakerSettings.InputGHEPFile: "/dune/data/users/cwret/tms/dune_nd_production_2022_v1/genie/00/neutrino.0_1649778420.ghep.root"
+#nd_cafmaker.CAFMakerSettings.NDLArRecoFile: "/dune/data/users/
+nd_cafmaker.CAFMakerSettings.TMSRecoFile: "/dune/app/users/cwret/SSRI/dune-ssri/neutrino.0_1649778420.edep_TMS_RecoCandidates_Hough_Cluster1.root"
```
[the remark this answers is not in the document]
clarenc3: Yup, have committed one of these example files now, thanks!

src/makeCAF.C
```diff
@@ -153,9 +154,12 @@ void loop(CAF& caf,
 
     caf.sr.run = par().runInfo().run();
     caf.sr.subrun = par().runInfo().subrun();
+    caf.sr.meta_run = par().runInfo().run();
+    caf.sr.meta_subrun = par().runInfo().subrun();
```
cjbackhouse, line 158: What's this?

clarenc3: Not my commit unfortuantely, no idea

src/reco/NDLArTMSMatchRecoFiller.h
clarenc3: Removed these

src/reco/NDLArTMSMatchRecoFiller.h
clarenc3: This has been addressed in latest commits

src/reco/TMSRecoBranchFiller.h
clarenc3: Kept these in for now. Would prefer to leave global variables out of it if we can, unless there's some benefit of having globals?

clarenc3:
Branch is now updated with a few additions:

1) Remove TF1 reliance and instead just use arithmetic with doubles
2) Update matching code to propagate the LAr track into the first layer of the TMS and do the residual on that point, instead of propagating both tracks into the dead region between the TMS and LAr. This removes reliance on the TMS unit vector for calculating the residual; the TMS unit vector can sometimes be unreliable because a straight line is a bad parameterisation of the track in the TMS due to the bending
3) Make TMS use cm as default unit for distances; LAr already does this and TMS was using mm. Also update the track matching code to know about this.

Found some bugs during this; the LAr direction vectors appear to be corrupted. Hence the matching calculates the line in LAr by using start and end points of the LAr track rather than the saved direction vector. This should be updated when the LAr unit vectors are fixed.

src/makeCAF.C
chenel: this was (still is?) the only way of copying run and subrun info in from the outside.

src/reco/TMSRecoBranchFiller.h
chenel: I think this is fine

src/reco/TMSRecoBranchFiller.cxx
chenel: let's put this on the "todo" list for the future.

chenel:
I think we are in a position to merge this (the ND-LAr bug above is not related to anything in this repo and will be fixed next time we respin).  @clarenc3  would you do the honors?

chenel:
Turns out @clarenc3 is ✈️ today, so I hereby invoke the merge. 🪄

Issue #8, opened 2022-08-10
Remove dumpTree requirement


clarenc3:
Currently need to run dumpTree before the CAF maker. This is a totally unnecessary step that could be baked into the CAF maker. Just a life enhancer, no new features

PR #10, opened 2022-08-18
Support filling SAND reco, remove dumpTree and parameterized reco

mvicenzi:
Three things in this pull request:
- removal of dumpTree dependency (solving #8)
- removal of parametrized reco
- addition of SAND reco fillers

Open issue on how to include SAND reco structures to open SAND reco file. We need a short term solution until SAND code is baked into a ups product. I'm currently loading manually our `.so` lib from a new parameter in the fhicl, but a better system is need to fix `#include SANDReco/SANDRecord.h` so that you don't need to put it in the path if you're not making a SAND CAF.
`SANDRecord.h` can be found in `/dune/app/users/mvicenzi/ND_CAFMaker/src/SANDReco/SANDReco/`.


cfg/ndcafmakerjob.fcl
```diff
@@ -2,7 +2,12 @@
...
-nd_cafmaker.CAFMakerSettings.InputDumpFile: "/dune/data/users/jwolcott/nd/nd-lar-reco/cafmaker_input/neutrino.LArTMSProductionJun23withLArCV.99.dump.root"
 nd_cafmaker.CAFMakerSettings.InputGHEPFile: "/dune/data/users/jwolcott/nd/nd-lar-reco/cafmaker_input/neutrino.LArTMSProductionJun23withLArCV.99.ghep.root"
 nd_cafmaker.CAFMakerSettings.OutputFile: "/dune/data/users/jwolcott/scratch/neutrino.LArTMSProductionJun23withLArCV.99.caf.root"
 nd_cafmaker.CAFMakerSettings.NDLArRecoFile: "/dune/data/users/jwolcott/nd/nd-lar-reco/cafmaker_input/neutrino.LArTMSProductionJun23withLArCV.99.summary.h5"
+
+# SAND test setup
+#nd_cafmaker.CAFMakerSettings.SANDRecoFile: "/pnfs/dune/scratch/users/mvicenzi/100k_FHC.0.fast-reco.root"
+#nd_cafmaker.CAFMakerSettings.SANDRecoLib: "/dune/app/users/mvicenzi/FastReco/lib/libStruct.so"
+#nd_cafmaker.CAFMakerSettings.InputGHEPFile: "/pnfs/dune/scratch/users/mvicenzi/100k_FHC.0.ghep.root"
+#nd_cafmaker.CAFMakerSettings.OutputFile: "/pnfs/dune/scratch/users/mvicenzi/test_SAND_CAF.root"
```
cjbackhouse, line 13: Can we maybe make a new `ndcafmakerjob_sand.fcl`?

src/makeCAF.C
```diff
@@ -198,31 +182,19 @@ int main( int argc, char const *argv[] )
...
 
   caf.version = 4;
   printf( "Run %d POT %g\n", caf.meta_run, caf.pot );
   caf.fillPOT();
 
-  // Copy geometric efficiency throws TTree to CAF file
-  std::cout << "Copying geometric efficiency throws TTree to output file" << std::endl;
-  TTree *tGeoEfficiencyThrowsOut = (TTree*) tf->Get("geoEffThrows");
-  caf.cafFile->cd();
-  tGeoEfficiencyThrowsOut->CloneTree()->Write();
```
cjbackhouse, line 224: What is this about? Is it something needed for PRISM?

mvicenzi: I believe so. @cmmarshall knows more. It was being copied over from the dumpTree, so in any case it needs to be reworked differently.

cjbackhouse: Confirmed verbally with Chris that we're OK to just delete this. It will need to be reimplemented for the full reco in any case.

src/Makefile
```diff
@@ -5,6 +5,19 @@ OBJS = $(patsubst %.cxx, %.o, $(SOURCES))
...
+# Automatically enable SAND if SANDRECO_INC or SANDRECO_LIB are set
+ifneq (x$(SANDRECO_INC), x)
+	INCLUDE += -I$(SANDRECO_INC)
+	SAND_OPT = -DENABLE_SAND
+endif
+
+ifneq (x$(SANDRECO_LIB), x)
+	LDLIBS += -L$(SANDRECO_LIB) -lStruct
+	SAND_OPT = -DENABLE_SAND
+endif
```
chenel, line 18: I think proper Make testing for an undefined variable looks like
```
ifeq ($(origin SANDRECO_INC), undefined)
...
endif
```
but I won't die on that hill

src/makeCAF.C
```diff
@@ -16,12 +16,11 @@
 #include "fhiclcpp/parse.h"
 
 #include "CAF.h"
-#include "dumpTree.h"
 #include "Params.h"
 #include "reco/MLNDLArRecoBranchFiller.h"
-#include "reco/ParameterizedRecoBranchFiller.h"
```
chenel, line 22: @cmmarshall wanted to keep the parameterized reco around for now, I thought

chenel (commented):
I see this has already been merged, but I want to check with @cmmarshall anyway about whether it's ok to drop the parameterized reco yet.  I didn't think that was the plan.

src/Makefile
cjbackhouse: What does `origin` do here? I'm very happy to learn something, but I had already saturated my tolerance for obscure conditional syntaxes trying to get the C preprocessor to do what I wanted...

chenel: > The result of this function is a string telling you how the variable variable was defined:
>
> ‘undefined’
>
>    if variable was never defined.
...
https://www.gnu.org/software/make/manual/html_node/Origin-Function.html

src/makeCAF.C
cjbackhouse: Confirmed with Chris. Getting rid of the parameterized reco is the plan. We can always rescue it from the history in case there's an unforseen need for it.

src/Makefile
cjbackhouse: Done and merged to master. It doesn't show up here apparently.

Issue #11, opened 2022-08-18
build is very fragile


chenel:
The build "system" is extremely fragile and requires a lot of hand-holding if anything ever changes in the environment.  We should perhaps contemplate switching to a CMake build to help manage dependencies and make it smoother for users.

PR #13, opened 2022-08-19
First pass at enabling FlatCAF making in ND

cjbackhouse:
I think this is approximately how it should look. Unfortunately it needs a newer C++/ROOT than is setup by default, and the ND_CAFMaker software setup is such a disaster that I haven't yet managed to get such a configuration to test.

There will also need to be some configuration options to enable/disable this, which I'm not sure how they should look.

### DIFF
--- a/src/CAF.cxx
+++ b/src/CAF.cxx
@@ -1,11 +1,13 @@
 #include "CAF.h"
 
+#include <regex>
+
 #include "duneanaobj/StandardRecord/Flat/FlatRecord.h"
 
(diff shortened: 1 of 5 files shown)


chenel:
Do you want to wait until after Luke has a whack at addressing #11?  He said he'll try it next week.

cjbackhouse:
I got this to build using the set of libraries in ndcaf_setup.sh, plus various other downstream changes. Doesn't run though.
Before merging this branch bfba4255f8c056679964ae9c99ed4cb8e296ec4a should probably be reverted (because hopefully something similar but better will be in use by then)

PR #15, opened 2022-08-25
fill `SRTrack::dir` for ND-LAr tracks

chenel:
The `SRTrack::dir` member was not being filled (`dir` was added after the last ND-LAr reco run).  This PR fixes that.

**This update will only work with `.h5` summary files produced using [commit 4b6965f of dune-nd-lar-reco](https://github.com/chenel/dune-nd-lar-reco/commit/4b6965fead19d00bf19344a155f0fd445076994a) or later, which does not include the first round of CAFs produced for the LBL workshop**


chenel:
Tested on sample file and outputs look sane.  Merging.

PR #17, opened 2022-10-20
Add skeleton of ND-GAr reco filler

fmartinezlopez:
Modified makeCAF.C and Params.h adding ND-GAr reco info variables. Created NDGArRecoBranchFiller able to fill some quantities in sr.nd.ndgar ("pseudo-reco" fields) using garsoft anatree.root.


chenel (commented):
looks fine to me!

PR #21, opened 2023-03-08
Sand variables

leadinoto:
Added SANDRecoFiller for filling reco variables from SAND-reco output file


chenel (changes requested):
Thanks for making this PR!  

The branch filling (in `SANDRecoBranchFiller`)  appears fine to me.  However, I don't think that adding  `SANDRecord.h` directly into the CAFMaker's repository is the right way to do use it.  I assume this header is copied from an equivalent SAND reconstruction package somewhere, right?  If so, what we need to do is `#include` it from there.  That way we don't mix dependencies.  (And I _especially_ don't want all the dictionary-related stuff at the bottom winding up in the CAFMaker library.)  There's already some code in `src/Makefile` that adds some SAND stuff into the build if the right environment variables are set, and I think that should be used to obtain this file.  Please remove the file from this branch.  Thanks!

mt82:
Given that this PR has been open since 2023 with no further activity and now has conflicts, I think it would make sense to close it

PR #23, opened 2023-05-06
Support reading "official" DeepLearnPhysics MLReco .h5 files

chenel:
Previously the CAFMaker was reading in a "stopgap" `.h5` format that was written out by me ages ago.  This PR implements support for `mlreco3d`'s "official" `.h5` outputs now being produced by the reco chain.


chenel:
I think this is mature enough to merge.  We'll make bugfixes as necessary

PR #28, opened 2023-09-30
Eliminate truth-matching hacks

chenel:
As of PicoRun4, `vertexID`s are no longer rewritten to be consecutive, and the 2x2-MLReco pipeline is fully propagating `vertexID`s through it, so we no longer need the workarounds that we previously needed.  Remove them here.

This comes with some new code to match run and `vertexID` numbers and updates to the storage classes that hold onto the GENIE tree(s) once loaded.

I also snuck in a few small updates (bypassing the GENIE splash screen, fixing a 'deprecated method' warning from FHICL).

This PR depends on some others that are used in the input file preparation:
https://github.com/DeepLearnPhysics/larcv2/pull/43
https://github.com/DeepLearnPhysics/SuperaAtomic/pull/17
https://github.com/DeepLearnPhysics/larnd2supera/pull/3

### DIFF
--- a/scripts/h5_to_cpp.py
+++ b/scripts/h5_to_cpp.py
@@ -16,6 +16,7 @@ numpy_to_hdf5 = {
     "i": "STD_I",
     "l": "STD_I",
     "B": "STD_U",
+    "L": "STD_I",
     "f": "IEEE_F",
     "d": "IEEE_F",
 }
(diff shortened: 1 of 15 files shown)


chenel:
Verified that this actually works on PicoRun4.1 files.  The `larnd2supera` PR listed above hasn't yet been merged, but we're going to merge to `flow2supera` very soon anyways, and the same commits will need to be ported there instead...

Issue #30, opened 2023-10-12
Possible kinematics calculation error in `FillTruth.cxx`


pgranger23:
While looking at `FillTruth.cxx` to adapt the FD CAFMaker to the new CAF format, I came across what seems to be a mistake in the momentum transfer calculation in `FillTruth.cxx`.

At `ND_CAFMaker/src/truth/FillTruth.cxx:147`, it seems that `lepP4` is defined but never set to its value before being used on L154, which impacts the later calculations.

Just reporting here as this seems wrong to me. Please excuse me if this has already been fixed or if this is the expected behaviour.

pgranger23:
I opened PR https://github.com/DUNE/ND_CAFMaker/pull/31 , with an attempt at fixing this. Just to make sure this doesn't get forgotten.

PR #31, opened 2023-10-18
Fix missing lepP4

pgranger23:
Just a small fix to address issue https://github.com/DUNE/ND_CAFMaker/issues/30
I did not have the opportunity to check it acts as expected on some input files.


chenel (approved):
thanks!

PR #39, opened 2023-11-13
Feature/minerva noe

noeroy:
MINERvA reco for NDCafMaker

Added the Minerva Branch Fille class and methods

Added Minerva into the parameter list and the CafMaker loop

### DIFF
--- a/src/makeCAF.C
+++ b/src/makeCAF.C
@@ -9,7 +9,6 @@
 #include "TFile.h"
 #include "TInterpreter.h"
 #include "TTree.h"
-#include <map>
 
 #include "cetlib/filepath_maker.h"
 #include "fhiclcpp/intermediate_table.h"
@@ -241,7 +240,7 @@ buildTriggerList(std::map<const cafmaker::IRecoBranchFiller*, std::deque<cafmake
     for (auto & fillerTrigPair : triggersByFiller)
     {
 
-      if (triggersByFiller.size() == 0) break;
+      //if (triggersByFiller.size() == 0) break;
       // we don't want to consider the stream we're already working with.
       // (but don't continue, because we want to remove this stream from the
       //  map if it's empty, per below)
--- a/src/makeCAF.C
+++ b/src/makeCAF.C
@@ -201,8 +201,12 @@ buildTriggerList(std::map<const cafmaker::IRecoBranchFiller*, std::deque<cafmake
 
   std::vector<std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>> ret;
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
   uto triggerTimeCmp = [](const cafmaker::Trigger & t1, const cafmaker::Trigger & t2)
+=======
+  auto triggerTimeCmp = [](const cafmaker::Trigger & t1, const cafmaker::Trigger & t2)
+>>>>>>> fixed last differences with MASTER
   {
     return t1.triggerTime_s < t2.triggerTime_s || (t1.triggerTime_s  == t2.triggerTime_s && t1.triggerTime_ns < t2.triggerTime_ns);
   };
@@ -290,11 +294,15 @@ void loop(CAF &caf,
   // figure out which triggers we need to loop over between the various reco fillers
   std::map<const cafmaker::IRecoBranchFiller*, std::deque<cafmaker::Trigger>> triggersByRBF;
   for (const std::unique_ptr<cafmaker::IRecoBranchFiller>& filler : recoFillers)
-  {
     triggersByRBF.insert({filler.get(), filler->GetTriggers()});
+<<<<<<< HEAD
   std::vector<std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>>
     groupedTriggers = buildTriggerList(triggersByRBF, par().cafmaker().trigMatchDT());
 
+=======
+  
+  std::vector<std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>> groupedTriggers = buildTriggerList(triggersByRBF);
+>>>>>>> fixed last differences with MASTER
   // sanity checks
   if (par().cafmaker().first() > static_cast<int>(groupedTriggers.size()))
   {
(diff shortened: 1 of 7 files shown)


src/reco/MINERvARecoBranchFiller.cxx
```diff
@@ -0,0 +1,508 @@
...
+    part.pdg = mc_traj_pdg[max_trkid];
+    part.G4ID = mc_traj_edepsim_trkid[max_trkid]; // Track id of Geant. reset for every interaction
+    part.interaction_id = mc_traj_edepsim_eventid[max_trkid]; // Vector ID of edepsim Unique number for all interactionss
+    part.time = mc_traj_pdg[max_trkid];
+    
+    part.time = mc_traj_point_t[max_trkid][0];
+    part.parent = mc_traj_parentid[max_trkid];
+
+    part.start_pos = caf::SRVector3D(mc_traj_point_x[max_trkid][0],mc_traj_point_y[max_trkid][0],mc_traj_point_z[max_trkid][0]);
+    part.end_pos = caf::SRVector3D(mc_traj_point_x[max_trkid][1],mc_traj_point_y[max_trkid][1],mc_traj_point_z[max_trkid][1]);
```
chenel, line 282: I think rather than creating a particle from scratch here you should instead be using the `TruthMatcher`'s `GetTrueParticle()` method, which works just like the `GetTrueInteraction()` one that you used below.  That will ensure you're never making duplicates, and can replace the 3-way branch you're doing below.  (see notes there)

src/reco/MINERvARecoBranchFiller.cxx
```diff
@@ -0,0 +1,508 @@
...
+          
+
+          sh.truth.ixn = truthVecIdx;
+          sh.truth.part = srTrueInt.nsec -1;
+          sh.truth.type = caf::TrueParticleID::kSecondary;
+        }
+      }
+    }
+
+  }
```
chenel, line 340: I'd really prefer you used the `TruthMatcher::GetTrueParticle()` method here rather than searching for / creating new particles explicitly.  If anything in the underlying implementation changes, this code could break unexpectedly, and it's better to have the "search for particle" code centralized in one place.  `GetTrueParticle()` should do everything you're doing here, and you can update the particle once it's returned to you if you need to.  (See the `ValidateOrCopy()` calls in the `MLNDLArRecoBranchFiller` for a clean way to do these updates.)

src/reco/MINERvARecoBranchFiller.cxx
```diff
@@ -0,0 +1,508 @@
...
+    caf::SRTrueParticle part;
+    part.pdg = mc_traj_pdg[max_trkid];
+    part.G4ID = mc_traj_edepsim_trkid[max_trkid]; // Track id of Geant. reset for every interaction
+    part.interaction_id = mc_traj_edepsim_eventid[max_trkid]; // Vector ID of edepsim Unique number for all interactionss
+    
+    part.time = mc_traj_point_t[max_trkid][0];
+    part.parent = mc_traj_parentid[max_trkid];
+
+    part.start_pos = caf::SRVector3D(mc_traj_point_x[max_trkid][0],mc_traj_point_y[max_trkid][0],mc_traj_point_z[max_trkid][0]);
+    part.end_pos = caf::SRVector3D(mc_traj_point_x[max_trkid][1],mc_traj_point_y[max_trkid][1],mc_traj_point_z[max_trkid][1]);
```
chenel, line 389: same comment as in the `track` version

src/reco/MINERvARecoBranchFiller.cxx
```diff
@@ -0,0 +1,508 @@
...
+          srTrueInt.sec.push_back(std::move(part));
+          srTrueInt.nsec ++;
+          
+
+          t.truth.ixn = truthVecIdx;
+          t.truth.part = srTrueInt.nsec -1;
+          t.truth.type = caf::TrueParticleID::kSecondary;
+        }
+      }
+    }
```
chenel, line 447: same comment as in the track version

chenel (changes requested):
Thank you so much for this!

The vast majority of it looks fine to me with only a few nitpicks.  The major thing I'm concerned about is the explicit loops through the true particle hierarchy, which I'd really like done instead via the dedicated `TruthMatcher::GetTrueParticle()` method in order to keep the logic in a single place.  See notes below.

src/makeCAF.C
```diff
@@ -209,6 +217,8 @@ buildTriggerList(std::map<const cafmaker::IRecoBranchFiller*, std::deque<cafmake
     const cafmaker::Trigger & trigSeed = trigGroup.front().second;
     for (auto & fillerTrigPair : triggersByFiller)
     {
+
+      //if (triggersByFiller.size() == 0) break;
```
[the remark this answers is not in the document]
noeroy: Fixed

src/reco/MINERvARecoBranchFiller.h
```diff
@@ -0,0 +1,144 @@
...
+      std::deque<Trigger> GetTriggers(int triggerType) const  override;
+
+
+      ~MINERvARecoBranchFiller() {
+        delete MnvRecoTree;
+        fMnvRecoFile->Close();
+        delete fMnvRecoFile;
+        MnvRecoTree = NULL;
+        fMnvRecoFile = NULL;
+      }
```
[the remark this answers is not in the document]
noeroy: Moved to `.cxx`

src/reco/MINERvARecoBranchFiller.cxx
noeroy: Followed the `MLNDLArRecoBranchFiller ` like implementation.
Had to "cheat" a bit to look for the primary, if we get a flag directly in Edepsim it will make things simpler.

Some Passthrough infomation are missing for to get the full `ValidateOrCopy()` working on things like position and momentum.

src/reco/MINERvARecoBranchFiller.cxx
noeroy: Same comment as in the `track` version too

chenel (approved):
Okay!  I'm happy with where things are at now, pending testing that we're actually still matching correctly after all the changes.  I'm happy to help debug further however if we still are having trouble not repeating particle creation using the `GetTrueParticle()` method in the truthmatcher.

PR #42, opened 2023-11-15
Bugfix/trigger match proposition

noeroy:
Specific changes I had on my attempt on trigger matching,
All of that will need to be more refined.


chenel (approved):
Looks good, let's just wait until we can test on updated MLreco files with fixed times in them.  Hopefully will have those this afternoon.

chenel:
Tested with most recent MLReco files in `/dune/data/users/jwolcott/nd-lar-reco/reco-out/PicoRun4.1/FixedTriggerTimes/` and seems to work ok.  We'll sort out any further issues on the branch for #39.

Issue #48, opened 2023-12-12
Fill rec.nd.dlp.track.truth and rec.nd.dlp.shower.truth correctly


sindhu-ku:
Tracks and Showers truth info need to be filled correctly since `std::vector<TrueParticleID> truth `is fixed in [duneanaobj ](https://github.com/DUNE/duneanaobj/issues/21)

PR #49, opened 2023-12-20
Bugfix/truth shw trk

sindhu-ku:
Filled truth branches of showers and tracks similar to particles in common branch (see [issue#48](https://github.com/DUNE/ND_CAFMaker/issues/48#issue-2038700553))


chenel (approved):
This is fine for now, assuming you've tested it to work.  But we should make an Issue for this so that we can extract all the copy-pasted code into a single function so that we don't duplicate it.  That will make for better maintenance long term.

PR #54, opened 2024-01-20
Pull truth vertex info from GENIE instead of ML-reco

chenel:
ML-reco's truth vertex locations have a known issue (in Supera particle edges are moved to the edge of the sensitive volume to aid in labeling for PPN, and the interaction vertex is inferred from those).  Stop using those.  Switch to GENIE instead.  

To verify: does this match MINERvA?


chenel:
This PR is superseded by #55 , where the MINERvA info is also included.

PR #55, opened 2024-01-24
Obtain true vertex info from GENIE files & compare to MINERvA

chenel:
ML-reco's truth vertex locations have a known issue (in Supera particle edges are moved to the edge of the sensitive volume to aid in labeling for PPN, and the interaction vertex is inferred from those). Stop using those. Switch to GENIE instead.  Also enable passing of true vertex info through MINERvA and ensure that matches GENIE.

Supersedes #54.


noeroy:
Added a commit to fix a trigger filling but happening with MINERvA

PR #56, opened 2024-02-09
uncommented truth GENIE vertex

noeroy:
Just uncommented the Genie filling of the vertex positions. You can see the difference before/after here. 
vtx_x_old
image

An interesting point is that we see that the vertex/neutrino filling is only done when "things" are reconstructed in any of the two detectors. the 1st event doesn't have any reconstructed objects in liquid argon and no showers/tracks in MINERvA, so no vertex is filled. Maybe we want to change that at some point?


chenel:
> An interesting point is that we see that the vertex/neutrino filling is only done when "things" are reconstructed in any of the two detectors. the 1st event doesn't have any reconstructed objects in liquid argon and no showers/tracks in MINERvA, so no vertex is filled. Maybe we want to change that at some point?

Yeah, we may want to, though this would involve each reco branch filler having its own call into the `TruthMatcher` for otherwise unreconstructed triggers.  Probably for that to work correctly we'd want to add a new pure virtual function to `IRecoBranchFiller` so that every reco branch filler would have to implement it.  But I'd need to think more about whether that's the right approach, so I think we should press on with this PR and come back to that.  I added #58 so we don't forget forever.

PR #59, opened 2024-03-04
Bugfix/minerva max traj

noeroy:
We store only `n_mc_trajectories` in the MINERvA dst file (10000 for now). In less than 1% spills, there are trajectories that go above that number. That lead to some CAFs making crashes for large number of CAFs.  Until we get a smarter way to store trajectories in the dst files, we just don't look for the true trajectory of that. specific track or shower and setting tthe type to kUnknown. 

linked are the error message we were getting for a specific file and the same run after the fix + the scan of that specific track that was creating an issue (with type == 0 now) 

bug_runnumber
bug_runnumber_fixed
bug_runnumber_type


chenel (approved):
looks fine to me.  thanks for the implementation!

PR #60, opened 2024-04-04
Added updated TMS Reco CAF fillers

LiamOS:
Full spill structure not yet accounted for, but it's also not fully ready on the TMS side either.
Includes relevant reco information included in StandardRecord 3 for TMS, no truth yet as I haven't had any GENIE files to work with; these should be available Soon™.

If you would like to test this I have some dune-tms reco output at `/exp/dune/data/users/losulliv/Jan2024_NERSC_MiniProd_Sample/tms-reco` in the RecoCandidates_Hough files, generated from the corresponding files in `../edep-sim` relative to previous.


chenel (approved):
thanks---looking forward to full integration 😄

Issue #62, opened 2024-04-24
Total POT not stored properly in flat cafs


mcasales:
Reading pot of file from` meta.pot` works well for cafs, but not for flatcafs. They still have the correct info in ` rec.beam.pulsepot`, but it seems the meta tree is filled incorrectly/empty. See output from opening flatcaf meta tree:
```
root -l /exp/dune/data/users/noeroy/prod/MiniRun5_1E19_RHC/MiniRun5_1E19_RHC.caf.beta1/CAF.flat/0000000/MiniRun5_1E19_RHC.caf.0000974.CAF.flat.root
root [5] meta->Scan("pot")
************************
*    Row   *       pot *
************************
Error in <TFile::ReadBuffer>: error reading all requested bytes from file /exp/dune/data/users/noeroy/prod/MiniRun5_1E19_RHC/MiniRun5_1E19_RHC.caf.beta1/CAF.flat/0000000/MiniRun5_1E19_RHC.caf.0000974.CAF.flat.root, got 0 of 307
Error in <TFile::ReadBuffer>: error reading all requested bytes from file /exp/dune/data/users/noeroy/prod/MiniRun5_1E19_RHC/MiniRun5_1E19_RHC.caf.beta1/CAF.flat/0000000/MiniRun5_1E19_RHC.caf.0000974.CAF.flat.root, got 0 of 78
Error in <TBranch::GetBasket>: File: /exp/dune/data/users/noeroy/prod/MiniRun5_1E19_RHC/MiniRun5_1E19_RHC.caf.beta1/CAF.flat/0000000/MiniRun5_1E19_RHC.caf.0000974.CAF.flat.root at byte:0, branch:pot, entry:0, badread=1, nerrors=1, basketnumber=0
*        0 *         0 *
************************
```
and output from meta tree in cafs:
```
root [1] meta->Scan("pot")
************************
*    Row   *       pot *
************************
*        0 *     1e+16 *
************************
```

noeroy:
It seems to be an issue with ROOT when we want to write the same tree with two opened files. I've added two minimal example scripts, one with the current way we fill the CAFs (test_bad.cpp) and one with the fix proposed in the branch iss62_flat_tree_fix.

After the fix we get:

[...]

PR #65, opened 2024-05-02
Bugfix: Issue 63 - Flat cafs trees not filled correctly

noeroy:
As discussed [in this issue](https://github.com/DUNE/ND_CAFMaker/issues/62), it seems to be an issue with ROOT when we want to write the same tree with two opened files. I've added two minimal example scripts, one with the current way we fill the CAFs (test_bad.cpp) and one with the fix I propose in this PR.

After the fix we get for the FLAT CAF:

``` 
root [0]
Attaching file test_cafs3.flat.root as _file0...
(TFile *) 0x29dd0e0
root [1] meta->Scan()
************************************************************
*    Row   *   pot.pot *   run.run * subrun.su * version.v *
************************************************************
*        0 * 1.015e+16 *         1 *         0 *         5 *
************************************************************
``` 
And for the CAF:
```
Attaching file test_cafs3.root as _file0...
(TFile *) 0x3a43120
root [1] meta->Scan()
************************************************************
*    Row   *   pot.pot *   run.run * subrun.su * version.v *
************************************************************
*        0 * 1.015e+16 *         1 *         0 *         5 *
************************************************************
```

[test.txt](https://github.com/DUNE/ND_CAFMaker/files/15183423/test.txt)
[test_bad.txt](https://github.com/DUNE/ND_CAFMaker/files/15183424/test_bad.txt)

### DIFF
--- a/src/CAF.cxx
+++ b/src/CAF.cxx
@@ -13,6 +13,8 @@ CAF::CAF(const std::string &filename, const std::string &rw_fhicl_filename, bool
   : pot(std::numeric_limits<decltype(pot)>::signaling_NaN()),  rh(rw_fhicl_filename)
 {
   cafFile = new TFile( filename.c_str(), "RECREATE" );
+  gDirectory->cd("Rint:/");
+
   cafSR = new TTree("cafTree", "cafTree");
   cafSRGlobal = new TTree("globalTree", "globalTree");
   cafMVA = new TTree("mvaTree", "mvaTree");
@@ -103,8 +105,8 @@ void CAF::write()
     {
       if (!tree)
         continue;
-
-      tree->Write();
+      auto clone_tree = (TTree*)tree->Clone();  
+      clone_tree->Write();
 
       // don't let it get stuck attached to only this file in case we need it again below
       tree->LoadBaskets();
--- a/src/CAF.cxx
+++ b/src/CAF.cxx
@@ -13,7 +13,6 @@ CAF::CAF(const std::string &filename, const std::string &rw_fhicl_filename, bool
   : pot(std::numeric_limits<decltype(pot)>::signaling_NaN()),  rh(rw_fhicl_filename)
 {
   cafFile = new TFile( filename.c_str(), "RECREATE" );
-  gDirectory->cd("Rint:/");
 
   cafSR = new TTree("cafTree", "cafTree");
   cafSRGlobal = new TTree("globalTree", "globalTree");
@@ -97,6 +96,20 @@ void CAF::fillPOT()
 
 void CAF::write()
 {
+
+  if(flatCAFFile){
+    flatCAFFile->cd();
+
+    for (auto tree : {cafSRGlobal, cafMVA, cafPOT, genie })
+    {
+      if (!tree)
+        continue;
+      auto clone_tree = (TTree*)tree->Clone();
+      clone_tree->Write();
+    }
+    flatCAFFile->Close();
+  }
+
   if(cafFile){
     cafFile->cd();
     cafSR->Write();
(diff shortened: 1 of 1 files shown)


src/CAF.cxx
```diff
@@ -13,6 +13,8 @@ CAF::CAF(const std::string &filename, const std::string &rw_fhicl_filename, bool
   : pot(std::numeric_limits<decltype(pot)>::signaling_NaN()),  rh(rw_fhicl_filename)
 {
   cafFile = new TFile( filename.c_str(), "RECREATE" );
+  gDirectory->cd("Rint:/");
+
```
chenel, line 17: Interesting!  I've never seen this before.  What does it do?

chenel (approved):
I don't see any obvious problems here, so let's move forward in the interest of things working for people.

src/CAF.cxx
noeroy: So, from my understanding of how things are created in ROOT, when you create a new file, you set the gDirectory to the new file (like when you cd). So if when you create the tree just after creating f1, the tree's directory will be f1. What I did was to reset the local directory to the base one ("Rint:/") so that the trees will be accessible by both files when we write the trees.  

 ```
 root [0] TFile* f1 = new TFile("f1.root", "RECREATE");
root [1]     TTree* t1 = new TTree("t1", "Tree for both files");
root [2] TFile* f2 = new TFile("f2.root", "RECREATE");
root [3] t1->GetDirectory()->GetName()
(const char *) "f1.root"
```
```
root [0] gDirectory->pwd()
Rint:/
root [1] TFile* f1 = new TFile("f1.root", "RECREATE");
root [2] gDirectory->cd("Rint:/");
root [3] TTree* t1 = new TTree("t1", "Tree for both files");
root [4] TFile* f2 = new TFile("f2.root", "RECREATE");
root [5] t1->GetDirectory()->GetName()
(const char *) "Rint"
```
In the minimal example I've sent, if you remove the gDirectory line, the tree doesn't get written properly in the second file. My understanding is that in that case, the trees are created in the structured CAF and not available for the flat CAFs. I just realized that another way to deal with that issue would have been to do something like that

```
Flat_file->cd()
For tree in trees();
  clone tree
  save the clone in the flat caf

then fill the structured caf as we do it
```

That way, we wouldn't have to deal with the Rint thing, as the clones are owned by the flat caf and the original by the structured. If you prefer it without it, I can implement it like that instead.

At least that's all what I understood from this outdated page and some tries and failures https://root.cern.ch/root/htmldoc/guides/users-guide/InputOutput.html

chenel: Maybe it would be better that way?  I'm worrying about future potential maintainers trying to reason out what this was doing and having trouble understanding it.  I think your proposed alternative is easier to understand, if it works...

noeroy: I'm also in favour of having the least "weird root specific commands" possible if we can avoid it. I'll implement my other solution, see if that works and keep you posted.

noeroy: Alright, I've adapted the code with the simpler method, and we still get the two meta trees filled properly.

The only thing that could be an issue is just memory wise, we're just going to store more information during the processing if we're cloning trees, but with the size of the information we're dealing with, I don't think it's going to affect us.

sindhu-ku: Regarding size: each module1 data file has ~90k events if that information changes anything

noeroy: The only things we clone are the last stage of the CAFs, I think that should be fine, as it's all smaller tries.

chenel: I agree, I doubt this will be a problem.

Issue #67, opened 2024-05-08
Propagate the parent and ancestor G4ID


sindhu-ku:
The track id of parent and ancestors are not filled with G4ID

Issue #66, opened 2024-05-02
Save G4 trajectory ID also for secondaries


sindhu-ku:
The G4 ID was not saved for secondaries initially at the flow stage and eventually also in the MLreco stage. Since this is now fixed in these two parts of the production, it is time to reflect that change also in ND_CAFMaker.

sindhu-ku:
The is_primary field needs to be fixed in the truth dataset in MLreco files in order to correctly propagate this information as the G4ID is used for the prim/sec distinction currently

Issue #64, opened 2024-04-29
True trajectory information does not match direction of momentum information


rdiurba:
Trajectories might be broken for particles as the momentum vector commonly has a different direction than the endpoints of the trajectory particle.

An example is this file: /exp/dune/data/users/noeroy/prod/MiniRun5_1E19_RHC/MiniRun5_1E19_RHC.caf.beta1/CAF/0000000/MiniRun5_1E19_RHC.caf.0000000.CAF.root

Here is an example in the file with indices to identify it:
160 (spill),0 (truth interaction index),0 (primary index),11.3667 (dX),0 (dY),0.886803 (dZ),0.975925 (direction relative to Z according to momentum info).

Example code (assuming you are already iterating over primaries in a truth interaction in a spill:
```auto start_pos=sr->mc.nu[ntrue].prim[primaries].start_pos;
auto end_pos=sr->mc.nu[ntrue].prim[primaries].end_pos;
auto p=sr->mc.nu[ntrue].prim[primaries].p;
double dX=abs(end_pos.x-start_pos.x);
double dY=abs(end_pos.y-start_pos.y);
double dZ=abs(end_pos.z-start_pos.z);
double length=TMath::Sqrt(dX*dX+dY*dY+dZ*dZ);
double lengthP=TMath::Sqrt(p.px*p.px+p.py*p.py+p.pz*p.pz);
if (sr->mc.nu[ntrue].iscc && abs(sr->mc.nu[ntrue].pdg)==14 && abs(pdg)==13){
std::cout<<n<<","<<ntrue<<","<<primaries<<","<<dX<<","<<dY<<","<<dZ<<","<<p.pz/lengthP<<std::endl;
```

sindhu-ku:
Can you check this first in the downstream MLrreco and flow files? This seems to be due to the known flow bug where the coordinates are interchanged for segments and trajectories.

noeroy:
I'll have a detailed look whenever I can, but is it possible that a particle scatters and keeps it's id, hence not going in a straight line and not being at the position of the projected p ?
I'll have a look at the edepsim file of that trajectory to see if that could make sense.

noeroy:
Alright, forget what I said earlier: 

I had a look at the true particle position with the several files of the chain (edepsim, flow, mlreco, minerva and cafs).

[...]

PR #68, opened 2024-05-10
Some minor MLreco changes for Minirun5 beta2 production

sindhu-ku:
1. MC truth start/end positions are now saved with position variables instead of start/end points ([issue #64](https://github.com/DUNE/ND_CAFMaker/issues/64)) 
2. True momentum is saved in MLreco files, so have added that to ValidateorCopy
3. Propagate the correct true `interaction_id` for true particles and ancestors
4. Propagate `genid ` for secondaries and determine if a particle is true primary in truth matching by looking for `genid` in the primary particle collection ([issue #66](https://github.com/DUNE/ND_CAFMaker/issues/66)) 
5. Parents and ancestors need to have their `genid `propagated instead of `trackid` for matching ([issue #67](https://github.com/DUNE/ND_CAFMaker/issues/67)). MLreco assigns the unique `trackid` as `parent_trackid` and `id` (unique within an event) as the `ancestor_trackid`. So this is done by looping over the particles to retrieve the `genid ` by matching the other two variables.

### DIFF
--- a/src/reco/MLNDLArRecoBranchFiller.cxx
+++ b/src/reco/MLNDLArRecoBranchFiller.cxx
@@ -228,9 +228,8 @@ namespace cafmaker
     };
 
     // todo: these need us to propagate nu info through Supera.  WIP
-//    ValidateOrCopy(ptTrueInt.nu_current_type, srTrueInt.iscc, false,
-//                   nuCurrComp, nuCurrSet);
-//    ValidateOrCopy(ptTrueInt.nu_energy_init, srTrueInt.E, NaN);
+    //ValidateOrCopy(ptTrueInt.nu_current_type, srTrueInt.iscc, false, nuCurrComp, nuCurrSet); //this is currently filled with -1 for iscc
+//    ValidateOrCopy(ptTrueInt.nu_energy_init/1000., srTrueInt.E, NaN); //this is currently filled with many -inf
 //    ValidateOrCopy(ptTrueInt.nu_interaction_mode, srTrueInt.mode, caf::ScatteringMode::kUnknownMode,
 //                   [](const NuInteractionMode & inCurr, const caf::ScatteringMode & outCurr)
 //                   {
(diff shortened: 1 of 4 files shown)


src/reco/MLNDLArRecoBranchFiller.cxx
```diff
@@ -532,8 +525,9 @@ namespace cafmaker
                                            std::find_if(sr.mc.nu.begin(),
                                                         sr.mc.nu.end(),
                                                         [&srTrueInt](const caf::SRTrueInteraction& ixn) {return ixn.id == srTrueInt.id;}));
+	  std::size_t truthPartIdx = std::distance(srTrueInt.prim.begin(), std::find_if(srTrueInt.prim.begin(), srTrueInt.prim.end(), [&srTrueInt, &truePartPassThrough](const caf::SRTrueParticle& part) { return part.G4ID == truePartPassThrough.gen_id; }));
+          bool is_primary = truthPartIdx != srTrueInt.prim.size();
```
chenel, line 529: this is a bit overcomplicated.  since you're using a `find_if()` anyway, you don't need the `distance()` (doing an extra unnecessary calculation).  think you could simplify to

```suggestion
    bool is_primary = std::find_if(srTrueInt.prim.begin(), srTrueInt.prim.end(), 
                                   [&srTrueInt, &truePartPassThrough](const caf::SRTrueParticle& part) { return part.G4ID == truePartPassThrough.gen_id; }) == srTrueInt.prim.end();
```

src/reco/MLNDLArRecoBranchFiller.cxx
```diff
@@ -644,8 +638,9 @@ namespace cafmaker
...
```
chenel, line 642: same here as above

```suggestion
    bool is_primary = std::find_if(srTrueInt.prim.begin(), srTrueInt.prim.end(), 
                                   [&srTrueInt, &truePartPassThrough](const caf::SRTrueParticle& part) { return part.G4ID == truePartPassThrough.gen_id; }) == srTrueInt.prim.end();
```

chenel (approved):
thanks for sorting all this out!  I noticed one small thing where the code is slightly more complicated than needs be, but otherwise, looks good!

src/reco/MLNDLArRecoBranchFiller.cxx
sindhu-ku: Thanks! Added it for showers as well. And it should be !=, not ==. So changed that

PR #73, opened 2024-08-02
Build with duneanaobj  v03_05_00

noeroy:
Updated setup script to build with newest version of duneanaobj

### DIFF
--- a/ndcaf_setup.sh
+++ b/ndcaf_setup.sh
@@ -4,15 +4,16 @@ setup gcc v9_3_0
 setup pycurl
 setup ifdhc
 setup geant4 v4_11_0_p01c -q e20:debug
-setup dk2nugenie   v01_10_01k -q debug:e20
+setup dk2nugenie   v01_10_01q -q e26:prof
 setup genie_xsec   v3_04_00 -q AR2320i00000:e1000:k250
 setup genie_phyopt v3_04_00 -q dkcharmtau
+#setup genie v3_04_02 -q debug:e26
 setup jobsub_client
 setup eigen v3_3_5
-setup duneanaobj v03_02_01 -q e20:prof
+setup duneanaobj v03_05_00 -q e20:prof
 setup hdf5 v1_10_5a -q e20
 setup fhiclcpp v4_15_03 -q debug:e20
-
+setup log4cpp v1_1_3e -q e26:prof
 
 # edep-sim needs to know where a certain GEANT .cmake file is...
 G4_cmake_file=`find ${GEANT4_FQ_DIR}/lib -name 'Geant4Config.cmake'`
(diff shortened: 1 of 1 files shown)


noeroy:
I'm able to make , run and the output files seems normal (can be opened, it doesn't complain and it's filled)

ndcaf_setup.sh
```diff
@@ -4,15 +4,15 @@ setup gcc v9_3_0
...
 setup genie_xsec   v3_04_00 -q AR2320i00000:e1000:k250
 setup genie_phyopt v3_04_00 -q dkcharmtau
 setup jobsub_client
 setup eigen v3_3_5
-setup duneanaobj v03_02_01 -q e20:prof
+setup duneanaobj v03_05_00 -q e20:prof
 setup hdf5 v1_10_5a -q e20
 setup fhiclcpp v4_15_03 -q debug:e20
-
+setup log4cpp v1_1_3e -q e26:prof
```
chenel, line 15: interesting that we need this now.  is this because we have a downstream dependency that lost it somehow?

chenel (approved):
thanks for making this work!!

ndcaf_setup.sh
noeroy: I'm not sure because it didn't seem to have anything specific to log4cpp in the include paths before, but the build complained about /cvmfs/larsoft.opensciencegrid.org/products/log4cpp/v1_1_3e/Linux64bit+3.10-2.17-e26-prof/lib/liblog4cpp.so: undefined reference to `std::__throw_bad_array_new_length()@GLIBCXX_3.4.29'

Now that I look a bit more into it, it seems that the error is not directly log4cpp, but there is a gcc conflict somewhere and that last step somewhat put us with the good compiler. Apparently we didn't need the dependency that did not have the right compiler, but I'll investigate more to make sure all the setup steps have the same gcc version.

noeroy: Alright, so digging into it, it raised some issue that you faced about 2 years from now with h5cpp.
Basically, on latestt duneanaobj, root version is v28, so for the Genie version to match it, we need to go from `e20` to `e26` in the build version.

We have e26 version of everything (yay) BUT the `e26` version of hdf5 doesn't have the h5cpp libraries in the include (I don't know why, could be that whomever built it didn't want to have the cpp side of it).

keeping just hdf5 as `e20` with the rest in `e26` works but it seems to be the kind of things that's going to bite us later?

chenel: With `e20` you're still able to read an mlreco file?  If so I say this is the best we can do.

noeroy: Yes, that still works fine, files are made as before, and no issues reading them. I'm merging.

PR #74, opened 2024-08-05
Changes for new SPINE (ML reco) format

sindhu-ku:
`particles` and `interactions` datasets have `reco` prefix now. Other name and type changes are automatically included in the DLP classes (includes new flash variables)

Manual name changes in ML reco branch filler:
1. `semantic_type` -> `shape`
2. `truth_id `-> `orig_id`
3. `truth_interaction_id` ->` orig_interaction_id`
4. `match` -> `match_ids`
5. `match_overlap` -> `match_overlaps`
6. No `Nu` before `InteractionMode` and `CurrentType`
7. No `truth_` before truth variables


chenel (approved):
thanks for keeping us up to speed!

PR #76, opened 2024-08-22
Add PandoraLArRecoNDBranchFiller

jback08:
Created the `PandoraLArRecoNDBranchFiller` class to store the reconstruction information from Pandora's [LArRecoND]( https://github.com/PandoraPFA/LArRecoND/tree/master) package (used for the DUNE ND). It requires a ROOT file created by the [HierarchyAnalysisAlgorithm](https://github.com/PandoraPFA/LArRecoND/blob/master/src/HierarchyAnalysisAlgorithm.cc#L141), which uses Pandora's [Hierarchy Tools](https://github.com/PandoraPFA/Documentation/blob/master/Hierarchy_Tools/Hierarchy_Tools_Overview.pdf).

The filled reco branches are `rec.nd.lar.pandora.tracks` and `.rec.nd.lar.pandora.showers`. No distinction is made (yet) between tracks and showers, and so the same 3D-cluster Particle Flow Objects (PFOs) are used for both. Here is the list of reco variables used:

1. Start position = cluster vertex point or the first hit position if no vertex is available
2. End position = cluster end position
3. Energy = [charge Q](https://github.com/DUNE/2x2_sim/wiki/File-data-definitions#chargecalib_prompt_hits)
4. Length = primary principal axis length
5. Quality = number of 3D hits.

Any changes needed here can be done mostly in the LArRecoND package, since the CAF just retrieves the output stored by Pandora's hierarchy algorithm.

The MC truth matching uses Pandora's Hierarchy Tools and not [TruthMatcher](https://github.com/DUNE/ND_CAFMaker/blob/main/src/truth/FillTruth.h#L126). Pandora requires **all MC particles to have a unique ID** (even if they originate from different neutrinos), and so the following offsets are applied to the original (ndlar-flow) MC Ids:

1. nuId = orig_nuId + 10^8
2. mcId = orig_mcId + nuIndex * 10^6

where nuIndex = 0 to N-1 for an event with N neutrinos, and the orig_mcId number range restarts from zero for each neutrino. This ensures all IDs are unique, for up to 100 neutrino interactions per event, each containing up to 10^6 hits.

These ID offsets are reversed when the Pandora CAF truth information is filled using the best reco-MC match achieved with Hierarchy Tools, and so the mcId's should then have values consistent with the equivalent input ndlar-flow files. The filled truth information corresponds to:

1. `truth.ixn` = orig_nuId
2. `truth.part` = orig_mcId (best match)
3. `truth.type` = primary, secondary or other
4. `truthOverlap` = completeness (best match)

H5 input files are first converted to ROOT using [h5_to_root_ndlarflow.py](https://github.com/PandoraPFA/LArRecoND/blob/master/ndlarflow/h5_to_root_ndlarflow.py)  before they are used by LArRecoND & Pandora, which in turn creates the hierarchy analysis output file that can be used to make the equivalent CAFs.

For the trigger, the event run numbers are propagated using those originally from the input ndar_flow ROOT files. Also, the start time is currently set using the [ts_start](https://github.com/DUNE/2x2_sim/wiki/File-data-definitions#chargeevents) variable, but this needs to be updated (in LArRecoND) to use the correct time and units.

Also updated `ndcaf_setup.sh` and `build.sh` to use a consistent environment.

### DIFF
--- a/src/reco/PandoraLArRecoNDBranchFiller.cxx
+++ b/src/reco/PandoraLArRecoNDBranchFiller.cxx
@@ -47,7 +47,6 @@ namespace cafmaker
 	  m_LArRecoNDTree->SetBranchAddress("dirZ", &m_dirZVect);
 	  m_LArRecoNDTree->SetBranchAddress("energy", &m_energyVect);
 	  m_LArRecoNDTree->SetBranchAddress("n3DHits", &m_n3DHitsVect);
-	  m_LArRecoNDTree->SetBranchAddress("length1", &m_length1Vect);
 	  m_LArRecoNDTree->SetBranchAddress("mcNuId", &m_mcNuIdVect);
 	  m_LArRecoNDTree->SetBranchAddress("isPrimary", &m_isPrimaryVect);
 	  m_LArRecoNDTree->SetBranchAddress("mcId", &m_mcIdVect);
@@ -118,7 +117,7 @@ namespace cafmaker
 	const float startZ = (m_startZVect != nullptr) ? (*m_startZVect)[i] : 0.0;
 	track.start = caf::SRVector3D(startX, startY, startZ);
 
-	// End position (length along principal direction or last hit location)
+	// End position
 	const float endX = (m_endXVect != nullptr) ? (*m_endXVect)[i] : 0.0;
 	const float endY = (m_endYVect != nullptr) ? (*m_endYVect)[i] : 0.0;
 	const float endZ = (m_endZVect != nullptr) ? (*m_endZVect)[i] : 0.0;
@@ -140,12 +139,14 @@ namespace cafmaker
 	const int n3DHits = (m_n3DHitsVect != nullptr) ? (*m_n3DHitsVect)[i] : 0;
 	track.qual = n3DHits*1.0;
 
-	// Cluster length along principal axis direction (cm)
-	const float length1 = (m_length1Vect != nullptr) ? (*m_length1Vect)[i] : 0.0;
-	track.len_cm = length1;
+	// Cluster length from start and end points
+	const float dX = endX - startX;
+	const float dY = endY - startY;
+	const float dZ = endZ - startZ;
+	track.len_cm = sqrt(dX*dX + dY*dY + dZ*dZ);
 
 	// Cluster length multiplied by LAr density (g/cm2)
-	track.len_gcm2 = length1*m_LArRho;
+	track.len_gcm2 = track.len_cm*m_LArRho;
 
 	// Use truth matching info from Pandora's LArContent hierarchy tools.
 	// For LArRecoND MC SpacePoints, we offset the MCId's to make them all unique:
--- a/src/reco/PandoraLArRecoNDBranchFiller.cxx
+++ b/src/reco/PandoraLArRecoNDBranchFiller.cxx
@@ -1,0 +1,306 @@
+#include "PandoraLArRecoNDBranchFiller.h"
+
+namespace cafmaker
+{
+
+  PandoraLArRecoNDBranchFiller::~PandoraLArRecoNDBranchFiller()
+  {
+      if (m_LArRecoNDFile && m_LArRecoNDFile->IsOpen())
+      {
+	  delete m_LArRecoNDTree; m_LArRecoNDTree = nullptr;
+      }
+      delete m_LArRecoNDFile; m_LArRecoNDFile = nullptr;      
+  }
+
+  PandoraLArRecoNDBranchFiller::PandoraLArRecoNDBranchFiller(const std::string &pandoraLArRecoNDFilename)
+  : IRecoBranchFiller("PandoraLArRecoND"),
+    m_Triggers(),
+    m_LastTriggerReqd(m_Triggers.end())
+  {
+      // Open Pandora LArRecoND hierarchy analysis ROOT file
+      m_LArRecoNDFile = TFile::Open(pandoraLArRecoNDFilename.c_str(), "read");
+      if (m_LArRecoNDFile && m_LArRecoNDFile->IsOpen()) {
+
+	  LOG.VERBOSE() << " Using PandoraLArRecoND file " << pandoraLArRecoNDFilename << "\n";
+	  
+	  // Input tree
+	  m_LArRecoNDTree = dynamic_cast<TTree*>(m_LArRecoNDFile->Get("LArRecoND"));
+	  if (!m_LArRecoNDTree) {
+	      LOG.FATAL() << "Did not find LArRecoND tree in input file " << pandoraLArRecoNDFilename << "\n";
+	      throw;
+	  }
+
+	  // Set branch addresses
+	  m_LArRecoNDTree->SetBranchAddress("event", &m_eventId);
+	  m_LArRecoNDTree->SetBranchAddress("run", &m_run);
+	  m_LArRecoNDTree->SetBranchAddress("subRun", &m_subRun);
+	  m_LArRecoNDTree->SetBranchAddress("startTime", &m_startTime);
+	  m_LArRecoNDTree->SetBranchAddress("sliceId", &m_sliceIdVect);
+	  m_LArRecoNDTree->SetBranchAddress("startX", &m_startXVect);
+	  m_LArRecoNDTree->SetBranchAddress("startY", &m_startYVect);
+	  m_LArRecoNDTree->SetBranchAddress("startZ", &m_startZVect);
+	  m_LArRecoNDTree->SetBranchAddress("endX", &m_endXVect);
+	  m_LArRecoNDTree->SetBranchAddress("endY", &m_endYVect);
+	  m_LArRecoNDTree->SetBranchAddress("endZ", &m_endZVect);
+	  m_LArRecoNDTree->SetBranchAddress("dirX", &m_dirXVect);
+	  m_LArRecoNDTree->SetBranchAddress("dirY", &m_dirYVect);
+	  m_LArRecoNDTree->SetBranchAddress("dirZ", &m_dirZVect);
+	  m_LArRecoNDTree->SetBranchAddress("energy", &m_energyVect);
+	  m_LArRecoNDTree->SetBranchAddress("n3DHits", &m_n3DHitsVect);
+	  m_LArRecoNDTree->SetBranchAddress("length1", &m_length1Vect);
+	  m_LArRecoNDTree->SetBranchAddress("mcNuId", &m_mcNuIdVect);
+	  m_LArRecoNDTree->SetBranchAddress("isPrimary", &m_isPrimaryVect);
+	  m_LArRecoNDTree->SetBranchAddress("mcId", &m_mcIdVect);
+	  m_LArRecoNDTree->SetBranchAddress("completeness", &m_completenessVect);
+	  
+	  // We have setup the input tree
+	  SetConfigured(true);
+      }
+  }
+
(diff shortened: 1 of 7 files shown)


noeroy:
I think that the truth part of it could be an issue, as the truth branches are shared by all reconstructions.

The particles are uniquely identified with two ids, one that identifies the interaction inside a given spill: `vertex_id` in the flow files, and one that is uniquely identifying for a given interaction which is the Geant4 trajectory id, I believe it is `traj_id` in the flow files. Those IDs are shared within MLReco and MINERvA so that truth particles are matched to the same id. Would it be possible to have a match between `orig_nuId `and `vertex_id` and between `orig_mcId` and `traj_id`?

noeroy:
I wonder if it wouldn't make more sense to only fill the tracks instead of filling twice the same information, especially for analysers that wouldn't know that for now track objects and shower objects here are the same thing. Also considering that it doesn't add any additional information right ?

jback08:
I think this is ready for review now. We have added the unix_ts variable to set the trigger time, and we are using the same MC Id variables as the other reco methods.

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -0,0 +1,317 @@
...
+	// nuIndex = int(mcId/10^6), so traj_id = mcId - nuIndex*10^6
+	const int mcNuId = (m_mcNuIdVect != nullptr) ? (*m_mcNuIdVect)[i] : 0;
+	const int vertex_id = (mcNuId > m_nuIdOffset) ? mcNuId - m_nuIdOffset : mcNuId;
+	const int isPrimary = (m_isPrimaryVect != nullptr) ? (*m_isPrimaryVect)[i] : -1;
+	const int mcId = (m_mcIdVect != nullptr) ? (*m_mcIdVect)[i] : 0;
+	const int nuIndex = int(mcId/m_maxMCId);
+	const int traj_id = mcId - nuIndex*m_maxMCId;
+	
+	caf::TrueParticleID trueID;
+	trueID.ixn = vertex_id;
```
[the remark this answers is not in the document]
noeroy: So `vertex_id` is defined [here](https://github.com/DUNE/2x2_sim/blob/develop/run-convert2h5/convert_edepsim_roottoh5.py#L331). That's the one in Flow hence the one used in Pandora if I'm not mistaken.  
The way I understand the code you sent me, it seems that it takes the 1st digit of the vertex ID, the last 5 digits and bunch them together. I'm not sure how you can recover the initial definition `vertex_id` from that. 

For current Minirun files, for instance run 15,  `TaggedRunId` varies between `15*10` and `16*10` for the neutrino events and `15*10+1e9` - `16*10 1e9` for rock muons. So basically the `run_number *10` annd `(run_number +1) *10`, and same +1e9 for rock muons.  The ratio 10 `TaggedRunId` per file is not set in stone though. 

For `EdepSimEventId`, for a MiniRun file, it can go up to 30 000 in what I've seen, but I wouldn't say it's an absolute limit.

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -0,0 +1,317 @@
...
+	if (isPrimary == 1) {
+	    trueID.type = caf::TrueParticleID::kPrimary;
+	} else if (isPrimary == -1) {
+	    trueID.type = caf::TrueParticleID::kUnknown;
+	} else {
+	    trueID.type = caf::TrueParticleID::kSecondary;
+	}
+	trueID.part = traj_id;
```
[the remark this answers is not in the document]
jback08: This has been updated as suggested.

src/reco/PandoraLArRecoNDBranchFiller.cxx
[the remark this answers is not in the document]
jback08: Some of the truth information is set using try-catch exception calls to catch those cases where no MC match is found (for secondaries with zero MC id's etc), avoiding code aborts.

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -0,0 +1,360 @@
...
+	// Use truth matching info from Pandora's LArContent hierarchy tools.
+	// For LArRecoND MC SpacePoints, we use the unique file_traj_id's
+	// for the MCParticles and the neutrino ID is the unique vertex_id.
+	// We also store the local traj_id's for the CAF truth matching tools
+
+	// Neutrino ID = vertex_id
+	const long mcNuId = (m_mcNuIdVect != nullptr) ? (*m_mcNuIdVect)[i] : 0;
+	// MC particle ID = traj_id
+	const long mcId = (m_mcLocalIdVect != nullptr) ? (*m_mcLocalIdVect)[i] : 0;
+	const int isPrimary = (m_isPrimaryVect != nullptr) ? (*m_isPrimaryVect)[i] : -1;
```
noeroy, line 264: Is that variable really the "Primary" definition of G4? Or a  variable that say if we have reconstructed a particle matched to a truth particle? 
If not we can use the id of the parent particle, if it's -1 then the particle is a Primary, otherwise it's a secondary.

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -0,0 +1,360 @@
...
+	} else if (isPrimary == -1) {
+	    truePartID.type = caf::TrueParticleID::kUnknown;
+	} else {
+	    truePartID.type = caf::TrueParticleID::kSecondary;
+	}
+
+	try
+	{
+	    // Get the true interaction in the stack
+	    caf::SRTrueInteraction &srTrueInt = truthMatch->GetTrueInteraction(sr, mcNuId);
```
noeroy, line 280: I'd avoid doing so if `mcNuID ==0` there are some failsafe, but then your logs look like that
```
[                                                          ]GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[                                                          ] 29m6s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[==>                                                       ] 3m5s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[====>                                                     ] 1m55s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[======>                                                   ] 1m18s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[=======>                                                  ] 1m15s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[=========>                                                ] 57s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[===========>                                              ] 50s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[=============>                                            ] 44s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
[==============>                                           ] 41s    GTreeContainer FATAL: Run number 0 was not found in this collection of .ghep files
```

and it looks like something went wrong for someone who wouldn't know the details. I think a simple if statment could avoid it

src/reco/PandoraLArRecoNDBranchFiller.cxx
jback08: This corresponds to the G4 definition of primary. This is only set to 1 when the particle parent is the neutrino (-1), otherwise it is set to 0. This uses the function [LArMCParticleHelper::IsPrimary()](https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArHelpers/LArMCParticleHelper.cc#L143) in Pandora.

noeroy: Alright, so if I read well the file, looks like secondaries are not associated to any true interaction (no vertex ID whenever `isPrimary==0`, this is not a CAFs issue and I think whenever that is implemented it wouldn't change what we're doing here.

jback08: The Pandora output is currently using the leading MCParticle for storing the truth matching info. I am finding out from the Pandora team if this is always set to be the corresponding primary particle. If so, this could explain why secondaries are not matched, and we will need to modify the Pandora output to include them as well. The updated secondary clusters output should then be automatically picked up by this CAF code.

src/reco/PandoraLArRecoNDBranchFiller.cxx
jback08: Changed to an if statement.

src/reco/PandoraLArRecoNDBranchFiller.cxx
jback08: There is a "dynamic folding" xml run parameter in Pandora's hierarchy tools that allows us to also store secondary particle information. The previous test files folded all MC matches to the primary particles, which meant that all of the entries had `isPrimary = 1` only. With the dynamic folding enabled, there should be `isPrimary = 0` entries appearing as well.

noeroy (approved):

src/reco/PandoraLArRecoNDBranchFiller.h
```diff
@@ -0,0 +1,82 @@
...
+      std::vector<float> *m_dirYVect = nullptr;
+      std::vector<float> *m_dirZVect = nullptr;
+      std::vector<float> *m_energyVect = nullptr;
+      std::vector<int> *m_n3DHitsVect = nullptr;
+      std::vector<long> *m_mcNuIdVect = nullptr;
+      std::vector<long> *m_mcLocalIdVect = nullptr;
+      std::vector<int> *m_isPrimaryVect = nullptr;
+      std::vector<float> *m_completenessVect = nullptr;
+
+      float m_LArRho{1.3973f}; // LAr density (g/cm3)
```
chenel, line 72: This might be better as a parameter, rather than hard-coded

chenel (approved):
Overall looks fine.  Just a couple of small grumbles.

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -0,0 +1,352 @@
...
+namespace cafmaker
+{
+
+  PandoraLArRecoNDBranchFiller::~PandoraLArRecoNDBranchFiller()
+  {
+      if (m_LArRecoNDFile && m_LArRecoNDFile->IsOpen())
+      {
+	  delete m_LArRecoNDTree; m_LArRecoNDTree = nullptr;
+      }
+      delete m_LArRecoNDFile; m_LArRecoNDFile = nullptr;
```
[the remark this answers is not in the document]
jback08: Changed to unique pointers.

src/reco/PandoraLArRecoNDBranchFiller.h
jback08: Added the LAr density as an optional config parameter with a default value.

Issue #72, opened 2024-08-02
Fill new flash objects


chenel:
Once we have a reco flash object in `StandardRecord` under `nd.lar` ([`duneanaobj` issue #40](https://github.com/DUNE/duneanaobj/issues/40)), we need to fill it.

PR #80, opened 2024-08-30
Fill flashes

sindhu-ku:
https://github.com/DUNE/ND_CAFMaker/issues/72
Changes to fill the new `SROpticalFlash `class and the reco-matched` FlashMatch` class in enums in `StandardRecord`
Tested using both latest data and simulation files (MR5) run through the whole updated ML chain


src/reco/MLNDLArRecoBranchFiller.o
chenel, line 1: another binary that I'm sure wasn't intended, let's remove

chenel (approved):
looks good apart from the two accidentally committed binary files!

PR #82, opened 2024-09-04
Fixed the trigger matching second to ns conversion

noeroy:
Fixed the s to ns conversion

### DIFF
--- a/src/makeCAF.C
+++ b/src/makeCAF.C
@@ -176,7 +176,8 @@ std::vector<std::unique_ptr<cafmaker::IRecoBranchFiller>> getRecoFillers(const c
 // -------------------------------------------------
 bool doTriggersMatch(const cafmaker::Trigger& t1, const cafmaker::Trigger& t2, unsigned int dT)
 {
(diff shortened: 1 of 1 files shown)


src/makeCAF.C
```diff
@@ -176,7 +176,8 @@ std::vector<std::unique_ptr<cafmaker::IRecoBranchFiller>> getRecoFillers(const c
 // -------------------------------------------------
 bool doTriggersMatch(const cafmaker::Trigger& t1, const cafmaker::Trigger& t2, unsigned int dT)
 {
-  return ( (std::max(t1.triggerTime_s, t2.triggerTime_s) - std::min(t1.triggerTime_s, t2.triggerTime_s)) * 10000000 + std::max(t1.triggerTime_ns, t2.triggerTime_ns) - std::min(t1.triggerTime_ns, t2.triggerTime_ns) ) < dT;
+  unsigned long int s_to_ns = 1e9;
```
chenel, line 179:
```suggestion
  const unsigned long int s_to_ns = 1e9;
```
let's make this `const`, since the compiler can do more optimization that way

chenel (approved):
thanks for fixing my stupid mistakes 🤦🏻

PR #86, opened 2024-09-16
Feature/track matching

noeroy:
Track matching algorithm added for 2x2 (spine and pandora) and Mx2.

Matching values and method are based on [DocDb 31970](https://docs.dunescience.org/cgi-bin/sso/ShowDocument?docid=31970). 

Modifications are: 
- MakeCaf -> Removed Matchers from the reco trigger branches (we don't store specific triggers for matchers)  and then called the matcher after all reco have filled their branches. 

- Matching is done in NDLArMINERvAMatchRecoFiller, with a matching in (Δθx, Δθy, Δx, Δy) done both for SPINE and Pandora (if one of them are missing no matching is done for it).

### DIFF
--- a/src/makeCAF.C
+++ b/src/makeCAF.C
@@ -168,7 +168,7 @@ std::vector<std::unique_ptr<cafmaker::IRecoBranchFiller>> getRecoFillers(const c
 
   if (!ndlarFile.empty() && !minervaFile.empty())
   {
-    recoFillers.emplace_back(std::make_unique<cafmaker::NDLArMINERvAMatchRecoFiller>(par().cafmaker().matchExtrapolatedZ(), par().cafmaker().matchdX(), par().cafmaker().matchdY(), par().cafmaker().matchdThetaX(), par().cafmaker().matchdThetaY()));
+    recoFillers.emplace_back(std::make_unique<cafmaker::NDLArMINERvAMatchRecoFiller>(par().cafmaker().trackMatchExtrapolatedZ(), par().cafmaker().trackMatchdX(), par().cafmaker().trackMatchdY(), par().cafmaker().trackMatchdThetaX(), par().cafmaker().trackMatchdThetaY()));
     std::cout << "   ND-LAr + MINERvA matching\n";
   } 
   // for now all the fillers get the same threshold.
--- a/src/makeCAF.C
+++ b/src/makeCAF.C
@@ -453,7 +453,10 @@ void loop(CAF &caf,
   // figure out which triggers we need to loop over between the various reco fillers
   std::map<const cafmaker::IRecoBranchFiller*, std::deque<cafmaker::Trigger>> triggersByRBF;
   for (const std::unique_ptr<cafmaker::IRecoBranchFiller>& filler : recoFillers)
+  {
+    if (filler->GetName() == "LArTMSMatcher" || filler->GetName() == "LArMINERvAMatcher") continue; //We don't want to store a trigger from a Matcher algorithm
     triggersByRBF.insert({filler.get(), filler->GetTriggers()});
+  }
   std::vector<std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>>
     groupedTriggers = buildTriggerList(triggersByRBF, par().cafmaker().trigMatchDT());
 
@@ -499,7 +502,6 @@ void loop(CAF &caf,
     {
       if (filler->GetName() == "LArTMSMatcher" || filler->GetName() == "LArMINERvAMatcher")
       {
-        std::cout<<filler->GetName()<<std::endl;
         filler->FillRecoBranches(groupedTriggers[ii][0].second, caf.sr, par, &truthMatcher);
       }
     }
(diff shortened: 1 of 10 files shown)


src/makeCAF.C
```diff
@@ -446,7 +453,10 @@ void loop(CAF &caf,
   // figure out which triggers we need to loop over between the various reco fillers
   std::map<const cafmaker::IRecoBranchFiller*, std::deque<cafmaker::Trigger>> triggersByRBF;
   for (const std::unique_ptr<cafmaker::IRecoBranchFiller>& filler : recoFillers)
+  {
+    if (filler->GetName() == "LArTMSMatcher" || filler->GetName() == "LArMINERvAMatcher") continue; //We don't want to store a trigger from a Matcher algorithm
```
chenel, line 457: Per the general comment, this would become

```suggestion
  {
    if (filler->FillerType != RecoFillerType::BaseReco) continue; //We don't want to store a trigger from a Matcher algorithm
```

src/makeCAF.C
```diff
@@ -486,14 +496,23 @@ void loop(CAF &caf,
       cafmaker::LOG_S("loop()").INFO() << "Global trigger idx : " << ii << ", reco filler: '" << fillerTrigPair.first->GetName() << "', reco trigger eventID: " << fillerTrigPair.second.evtID << "\n";
       fillerTrigPair.first->FillRecoBranches(fillerTrigPair.second, caf.sr, par, &truthMatcher);
     }
+
+    // Once all the reco fillers have been called, let's call the matching fillers
+    for (const std::unique_ptr<cafmaker::IRecoBranchFiller>& filler : recoFillers)
+    {
+      if (filler->GetName() == "LArTMSMatcher" || filler->GetName() == "LArMINERvAMatcher")
```
chenel, line 503: and this one would become

```suggestion
      if (filler->FillerType == RecoFillerType::Matcher)
```

src/reco/NDLArMINERvAMatchRecoFiller.cxx
```diff
@@ -0,0 +1,209 @@
...
+
+  bool NDLArMINERvAMatchRecoFiller::Passes_cut(caf::SRTrack track_minerva, caf::SRTrack track_Lar, double &costheta, double &residual) const
+  {
+
+    // HARDCODED VALUES BASED ON DOCDB 31970
+    double z_extr = -70;   // Extrapolated position compariton. Here it's the front of the Lar modules.
+    double d_x = 17;       // Maximum residual in x coordinate [cm];
+    double d_y = 19;       // Maximum residual in y coordinate [cm];
+    double d_thetax = .08; // Maximum Angle difference wrt to x axis [rad];
+    double d_thetay = .09; // Maximum Angle difference wrt to y axis [rad];
```
chenel, line 38: Can we please make these configuration parameters instead of hard-coding??

src/reco/NDLArMINERvAMatchRecoFiller.h
```diff
@@ -0,0 +1,35 @@
+/// Fill ND-LAr reco branches using DeepLearnPhysics machine learning based reconstruction.
+///
+/// \author  J. Wolcott <jwolcott@fnal.gov> & F. Akbar <fakbar@ur.rochester.edu>
```
chenel, line 3: Need to update the copy-paste 😄

chenel (changes requested):
Hmm.  I don't love the filler name string comparisons in this code; they're rather fragile and easily lost if somebody happens to change the name later.  I think a better approach would be to add a new required method to the `IRecoBranchFiller` interface called something like `FillerType` that returns a member of a new enum that we can use to differentiate them (obviously better names welcome---I came up with these with <30s thought):

```
enum class RecoFillerType
{
   Unknown,
   BaseReco,  ///<  Full reconstruction stack like SPINE or Pandora
   Matcher,     ///<  Post-hoc matching run across detectors, but not creating new trigger entries etc.
};
```

and then adding

```
  /// What type of IRecoBranchFiller is this?
  RecoFillerType  Filler Type() const = 0;
```

to the `IRecoBranchFiller` class.  Then the `Pandora`, `SPINE`, `MINERvA` recos will override it with

```
   RecoFillerType FillerType() const override { return RecoFillerType::BaseReco; }
```

while the matchers will instead return

```
   RecoFillerType FillerType() const override { return RecoFillerType::Matcher; }
```

Then the "type" of the filler is embedded directly in the code, which is a bit more maintainable and obvious, and harder to break accidentally.  I'll suggest the resulting changes to the code in this PR in line-by-line comments.

src/makeCAF.C
noeroy: Implemented

src/makeCAF.C
noeroy: Implemented

src/reco/NDLArMINERvAMatchRecoFiller.h
noeroy: Fixed it

src/reco/NDLArMINERvAMatchRecoFiller.cxx
noeroy: Done, when TMS will do track matching we might want to have some common things, until now I've just created `TrackMatch` variables to store all of those and explained the default value from the docdb entry.

chenel (approved):
Looks good.  Thanks, as always, for humoring my pedantry 😄

noeroy:
Thanks for reviewing!

PR #89, opened 2024-10-11
Interface for IFbeam

sindhu-ku:
- Added new class for IFbeam interface in src/beam
- Can get POT info directly from IFBeam
- Have to also query other devices to get other beam info like horn current, position, etc.,


src/beam/IFBeam.h
```diff
@@ -0,0 +1,34 @@
...
+#include <map>
+#include <vector>
+#include <nlohmann/json.hpp>
+
+#include "Params.h"
+#include "CAF.h"
+#include "reco/IRecoBranchFiller.h"
+
+using BeamSpills = std::map<double, double>;
+using json = nlohmann::json;
```
chenel, line 14: `using` declarations in the global namespace in a header file is Bad Form, because any subsequent file that either `#include`s this header or is compiled in the same compilation unit as this header will have the aliases forced on them---which can cause accidental collisions.  In this case:

* `BeamSpills` is only used in a single place in the header, and all its occurrences both here and in the `.cxx` are within the class scope.  Just move this `using` declaration inside the `public` block of `IFBeam` and it should be fine.  (This avoids any collisions because the alias is then `IFBeam::BeamSpills`, which will never overlap anything else.)
* The `json` alias is only used in the `.cxx`, so move the `using` declaration there.  (Then the alias only exists when that `.cxx` is being compiled.)

src/beam/IFBeam.h
```diff
@@ -0,0 +1,34 @@
...
+public:
+    IFBeam(const std::vector<std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>>& groupedTriggers, bool is_data);
+    
+    double getPOT(const cafmaker::Params& par, std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>& groupedTrigger, int ii);
+
+private:
+    const std::string potDevice = "E:TRTGTD";
+    BeamSpills beamSpills;
+
+    void loadBeamSpills(const std::vector<std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>>& groupedTriggers);
```
chenel, line 26: While we're making aliases, adding 
```
using TriggerGroup = std::vector<std::pair<const cafmaker::IRecoBranchFiller*, cafmaker::Trigger>>;
```
to the `public` block of the class would simplify these a lot:
```suggestion
    IFBeam(const std::vector<std::vector<TriggerGroup>& groupedTriggers, bool is_data);
    
    double getPOT(const cafmaker::Params& par, TriggerGroup & groupedTrigger, int ii);

private:
    const std::string potDevice = "E:TRTGTD";
    BeamSpills beamSpills;

    void loadBeamSpills(const std::vector<TriggerGroup>& groupedTriggers);
```

src/beam/IFBeam.h
```diff
@@ -0,0 +1,34 @@
...
+    std::string createUrl(const std::string& min_time_iso, const std::string& max_time_iso);
+    double unitToFactor(const std::string& unit);
+};
+ 
+std::string toISO8601(double time_sec);
+double getTriggerTime(const cafmaker::Trigger& trigger);
```
chenel, line 32: These are also in the global namespace, and I don't think they really should be.  In fact I think the whole file should probably be inside `namespace cafmaker`, and it might even be worth moving these functions to the `util` subdirectory and making them further inside `cafmaker::util`...

src/beam/IFBeam.cxx
```diff
@@ -0,0 +1,172 @@
+#include "util/Logger.h"
+#include "util/Progress.h"
+#include "IFBeam.h"
+#include <curl/curl.h>
+#include <chrono>
+#include <regex>
+#include <sstream>
+#include <iomanip>
+#include <limits>
+
```
chenel, line 10: If you take my suggestion on the `.h`, either a `using namespace cafmaker` or (better, I think) enclosing the whole file from here in `namespace cafmaker { }` will be necessary

src/beam/IFBeam.cxx
```diff
@@ -0,0 +1,172 @@
...
+    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(tp.time_since_epoch()) % 1000;
+    if (ms.count() > 0) {
+        ss << '.' << std::setfill('0') << std::setw(3) << ms.count();
+    }
+    ss << (std::localtime(&in_time_t)->tm_gmtoff >= 0 ? '+' : '-')
+       << std::setfill('0') << std::setw(2) << std::abs(std::localtime(&in_time_t)->tm_gmtoff) / 3600
+       << ':' << std::setfill('0') << std::setw(2) << (std::abs(std::localtime(&in_time_t)->tm_gmtoff) % 3600) / 60;
+
+    return ss.str();
+}
```
chenel, line 37: as noted in the `.h`, I think there's a good argument that these don't belong in `IFBeam.cxx` (they're not IFBeam-specific)---they might be better served in a `util/db_utils` or something like that.  Worth considering at least.

src/beam/IFBeam.cxx
```diff
@@ -0,0 +1,172 @@
...
+        try {
+            auto json_data = json::parse(readBuffer);
+            for (const auto& spill : json_data["rows"]) {
+                double time = spill["clock"].get<double>() * ms_to_s;
+                std::string unit = spill["units"];
+                double pot = spill["value"].get<double>() * unitToFactor(unit);
+                beamSpills[time] = pot;
+            }
+        } catch (const json::exception& e) {
+            std::cerr << "Failed to parse JSON data: " << e.what() << std::endl;
```
chenel, line 109: do we really want to just write an error message and try to continue?  the result will be garbage POT values, which will take more effort to track down later.  I think the default behavior should be to crash right away.  If we _want_ to give people the ability to continue without POT---and I'd be interested to hear an argument for this use case, because I'm not seeing it offhand---I think it should be a FCL option that defaults to `false`.

chenel (approved):
The overall approach here looks sound to me.  I've made some comments inline about the strategy for when the IFBeam lookup fails and about the coding style---once we resolve those, we'll be good to go.

src/Params.h
```diff
@@ -45,12 +45,12 @@ namespace cafmaker
...
     fhicl::OptionalAtom<std::string> minervaRecoFile  { fhicl::Name{"MINERVARecoFile"}, fhicl::Comment("Input MINERVA reco .root file") };
 
-    // fixme: temporary hack for data.  should be removed when interface to IFDB is complete
-    fhicl::OptionalAtom<std::string> POTFile { fhicl::Name{"POTFile"}, fhicl::Comment("Input txt file with beam spill information") };
     
     // this is optional by way of the default value. Will result in an extra output file if enabled
     fhicl::Atom<bool> makeFlatCAF { fhicl::Name{"MakeFlatCAF"}, fhicl::Comment("Make 'flat' CAF in addition to structured CAF?"), true };
 
+    fhicl::Atom<bool> IsData { fhicl::Name("IsData"), fhicl::Comment("True if data"), false}; //Get IFbeam info if true, default is false
+
```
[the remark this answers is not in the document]
sindhu-ku: I am not sure. I was first thinking about basing it on whether there are GENIE or edep sim files given as inputs. But, sometimes maybe we wan't to run the simulation w/o truth? I am not sure why that would be the case, but we have done it in the past when truth matching had issues.

ndcaf_setup.sh
```diff
@@ -14,7 +14,9 @@ setup hdf5 v1_10_5a -q e20
 setup fhiclcpp v4_15_03 -q debug:e20
 setup edepsim v3_2_0c -q debug:e20
 
-# edep-sim needs to know where a certain GEANT .cmake file is...
+export LD_LIBRARY_PATH=$CURL_ROOT/lib:$LD_LIBRARY_PATH
```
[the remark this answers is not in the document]
sindhu-ku: I couldn't find it in the dune products folder. Is this an issue
Screenshot from 2024-10-17 14-12-54

src/beam/IFBeam.cxx
sindhu-ku: You are right, I missed this. Added abort also when the https request doesn't return anything.

sindhu-ku:
> The overall approach here looks sound to me. I've made some comments inline about the strategy for when the IFBeam lookup fails and about the coding style---once we resolve those, we'll be good to go.

Ok I think I have included all of your suggestions. Just need a smarter way to recognize a data file instead of using Params

src/Params.h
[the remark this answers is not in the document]
chenel: Okay.  Maybe then the thing to do here is:

- Use presence of (GENIE || edep-sim) as the default decision as to whether to use IFBEAM
- Change this switch's name from `IsData` to `ForceDisableIFBeam` (or whatever you want to call it), since that's actually the behavior this switch is meant to control.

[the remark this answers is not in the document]
sindhu-ku: Makes sense, updated it accordingly.

chenel:
I wonder if you've been waiting for me?  I think we should go ahead with what we have here for now, and continue to look for a way to determine whether a file is data in the future

sindhu-ku:
Oops didn't realize it was approved. Merging now

PR #91, opened 2024-11-14
Update SPINE format for MiniRun 6.2

sindhu-ku:
Changes:

1. We no longer need `gen_id` since this info is now saved in `track_id` (hope I don't have nightmares about this anymore)
2. There are also other variable name/type changes in SPINE. The ones that affect CAFs are matched flash variables. A reco interaction can have more than one matched flash, so the variable types and names have to be changed also in the `StandardRecord`. For now, I am not filling the id and time variables of the matched flash(es) until we make that change. We don't have to worry about this for now since there is no flash matching. Hope to fill this after MiniRun 6.2 so we enough time to also request a new `duneanaobj` version

### DIFF
--- a/src/reco/DLP_h5_classes.cxx
+++ b/src/reco/DLP_h5_classes.cxx
@@ -5,7 +5,7 @@
 //
 //    The invocation that generated this file was:
 //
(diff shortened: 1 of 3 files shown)


chenel (approved):
looks great!

PR #92, opened 2024-11-15
Bugfix for trigger matching

noeroy:
When 3 trigger streams are used as input, there seem to be a memory issue in the trigger seed.

I assume it's because the seed was updated at each time the  `trigGroup` was updated with only two streams compared to now where we can still have the same seed while the `trigGroup` is updated.

Joined 2 images before and after the change with a `cout ` to show the trigger seed (or trigGroup.front().second) time + the number of matched triggers.

Before the fix, we had some crazy times for the seed and additional 19 triggers due to that issue that is fixed by that commit. (see attached [fix_trigger.pdf](https://github.com/user-attachments/files/17781421/fix_trigger.pdf) )


chenel (approved):
Hmm.  This seems to be a problem with the `trigSeed` reference suddenly not referring to anything and having junk memory in it.  I can't see _how_ based on the code; `trigGroup` gets updated, but only at the _back_ of the dequeue, so the front should still be ok.  Oh well, if this works empirically and fixes the memory stomping, then let's do it.

PR #93, opened 2025-01-13
Add Pandora interactions to the standard record

jback08:
This adds the interaction information to the common part of the standard record for Pandora, which is currently missing in the CAFs. Each `SRInteraction` contains the reconstructed neutrino vertex along with the tracks and showers that are retrieved from the Pandora reco record, which are filled before the interactions.

We also fix an issue with the size of `sr.nd.lar.pandora`, which stores the reco tracks and showers, that was incorrectly set to the number of clusters and not the number of neutrinos.


chenel (commented):
I don't see any red flags here.  However I'd appreciate a review from @noeroy as well because he was following this more closely.

noeroy (approved):
Seems good to me,
quickly tested it comparing with the other branch and everything seem table (tracks & truth link)

PR #95, opened 2025-03-06
Use reco not MC PDG hypothesis & isPrimary boolean for Pandora CAFs

jback08:
Update the PDG hypothesis & isPrimary boolean to use the reconstruction not MC truth info for Pandora. 

Also moved the filling of the interaction information from `FillInteractions()` (which is now removed) to the `FillTracks()` and `FillShowers()` functions, depending if the Pandora clusters are tracks or showers, respectively. 

The input Pandora ROOT file stores the reco PDG hypothesis & isPrimary variables for each cluster, and it is easier to set these interaction variables when the track & shower Standard Record objects are being created, rather than retrieving them from the Standard Record, as was done by the previous `FillInteractions()` function, and then trying to work out which Pandora cluster ntuple entry they should correspond to (e.g. using a map).

This update requires LArRecoND [v01-01-05](https://github.com/PandoraPFA/LArRecoND/tree/v01-01-05) for adding the two new reco ROOT variables `isRecoPrimary` and `recoPDG`, as well as LArContent [v04_14_00](https://github.com/PandoraPFA/LArContent/tree/v04_14_00) for adding functionality to say whether a reconstructed cluster is a primary one or not.

### DIFF
--- a/src/reco/PandoraLArRecoNDBranchFiller.cxx
+++ b/src/reco/PandoraLArRecoNDBranchFiller.cxx
@@ -52,6 +52,8 @@ namespace cafmaker
 	  m_LArRecoNDTree->SetBranchAddress("nuVtxX", &m_nuVtxXVect);
 	  m_LArRecoNDTree->SetBranchAddress("nuVtxY", &m_nuVtxYVect);
 	  m_LArRecoNDTree->SetBranchAddress("nuVtxZ", &m_nuVtxZVect);
+	  m_LArRecoNDTree->SetBranchAddress("isRecoPrimary", &m_isRecoPrimaryVect);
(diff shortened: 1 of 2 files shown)


chenel (approved):
hey John: I don't have the bandwidth to give these careful reviews like Noë was doing, but this looks ok to me on a quick read-through.  If you need/want a more detailed review, we can see if @noeroy has the cycles for it 😄

noeroy (approved):
I don't see much issue with it. 
Can I assume you tried to run and it gave sensible results?

jback08:
Yes, I have ran this on an example MiniRun6.2 file and the results look OK.

PR #96, opened 2025-04-03
added minerva data processing specific features

noeroy:
- Offset between Mx2 software and 2x2 ref point is not passed in the data processing, so it has to be hardcoded for now for data (branches value don't make any sense). Will make an issue as long as it's pushed to remind us that we don't want the offset to be hardcoded here but passed inside the MINERvA file.
 - Minerva meta "event" now points to the entry in the Minerva input tree, it'll make life easier for backtracking in the dst file.
 - Trigger is not shifted back to 0 for data (otherwise the code would crash and the matching wouldn't make sense)

Tested for data files and thing do make sense (track positions are where we expect them to be + triggers are matched as expected).
Only touches data (`ev_gps_time_sec>1.5e9`, for MC it would be 0(0) ) so shouldn't impact any MC production.

### DIFF
--- a/src/reco/MINERvARecoBranchFiller.cxx
+++ b/src/reco/MINERvARecoBranchFiller.cxx
@@ -29,10 +29,7 @@ namespace cafmaker
         throw;
       }
 
-      //Meta branches
-      MnvRecoTree->SetBranchAddress("offsetX", &offsetX);
-      MnvRecoTree->SetBranchAddress("offsetY", &offsetY);
-      MnvRecoTree->SetBranchAddress("offsetZ", &offsetZ);
+     
 
       MnvRecoTree->SetBranchAddress("ev_trigger_type", &ev_trigger_type);
       MnvRecoTree->SetBranchAddress("ev_gl_gate", &ev_gl_gate);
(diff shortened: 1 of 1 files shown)


sindhu-ku (approved):

src/reco/MINERvARecoBranchFiller.cxx
```diff
@@ -118,6 +115,20 @@ namespace cafmaker
       MnvRecoTree->GetEntry(0);
       is_data =  ev_gps_time_sec>1.5e9; 
 
+      //Meta branches
+      if (is_data)
+      {
+          offsetX = 0;
+          offsetY = 218;
+          offsetZ = 6913;
+      }
```
chenel, line 124: Please don't hard-code numbers in the CAFMaker.  These should be parameters that live in the .fcl.

chenel (changes requested):
Fine by me, apart from the hard-coding of the offsets.  Let's convert them into parameters that can be in the FCL, and then I'm all good.

src/reco/MLNDLArRecoBranchFiller.cxx
```diff
@@ -797,7 +797,7 @@ namespace cafmaker
       // fill flash variables for all flashes
 
       opflash.id = flash.id;
-      opflash.tpc_id = flash.tpc; //TODO
+      //opflash.tpc_id = flash.tpc; //TODO
```
sindhu-ku: It should be flash.volume_id
That's why it was giving you error. Sorry missed it

src/reco/MINERvARecoBranchFiller.cxx
noeroy: Done

chenel (approved):
thanks!

PR #98, opened 2025-05-01
Set Pandora trigger type.

jback08:
The trigger type integer "triggers" is now stored in the Pandora hierarchy ROOT output file in LArRecoND tag [v01-02-00](https://github.com/PandoraPFA/LArRecoND/tree/v01-02-00), which includes the code added in [PR29](https://github.com/PandoraPFA/LArRecoND/pull/29) and [PR30](https://github.com/PandoraPFA/LArRecoND/pull/30). This integer is now used to set the trigger type for the Pandora CAFs.

### DIFF
--- a/src/reco/PandoraLArRecoNDBranchFiller.cxx
+++ b/src/reco/PandoraLArRecoNDBranchFiller.cxx
@@ -32,6 +32,7 @@ namespace cafmaker
 	  m_LArRecoNDTree->SetBranchAddress("subRun", &m_subRun);
 	  m_LArRecoNDTree->SetBranchAddress("unixTime", &m_unixTime);
 	  m_LArRecoNDTree->SetBranchAddress("startTime", &m_startTime);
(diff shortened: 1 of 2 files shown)


noeroy:
Hi, thanks for that work, 
What numbers will we expect for pandora? Is it the IO group or a default integer if there's no external triggers?

jback08:
It uses the iogroup:

```
triggers = flow_out["charge/events","charge/ext_trigs",events["id"][ievt]]
triggerArray=triggers["iogroup"]
triggerID = np.array(np.ma.getdata(triggerArray),dtype='int32')
```

and this triggerID is then stored as `triggers` in the Pandora ROOT output.

noeroy:
And how do you do for the event where there is no ext_trigs?  Do you have a test file somewhere that I could x-check quickly?

jback08:
We have made LArRecoND tag [v01-02-01](https://github.com/PandoraPFA/LArRecoND/tree/v01-02-01) which includes the trigger info used for this PR.

Issue #97, opened 2025-04-11
Treatment of non beam events in CAFMaker


noeroy:
CAFMaker is currently thought for only beam events, when filling non beam events we get some "fake non 0 POT" addition and the Mx2 matching is messed up.

To see that effect, I've made a comparison of two files taken on the same time frame during summer 2024:

```
/global/cfs/cdirs/dune/www/data/2x2/sandbox/v6/CAFs.beam_only/with_pandora/CAF.flat/packet-0050017-2024_07_09_00_14_34_CDT.CAF.flat.root
(beam only)


root [1] cafTree->Draw("rec.beam.pulsepot>>htemp","rec.beam.pulsepot>0","goff")
(long long) 460
root [2] cout<<htemp->GetMean() * htemp->GetEntries()<<endl;
2.33712e+16
root [3] cafTree->GetEntries("rec.nd.minerva.nixn>0 && rec.nd.lar.ndlp>0 ")
(long long) 305
```

```
/global/cfs/cdirs/dune/www/data/2x2/reflows/v9/caf.flat/beam/july8_2024/nominal_hv/packet-0050017-2024_07_09_00_14_34_CDT.CAF.flat.root 
(beam + non beam)

root [1] cafTree->Draw("rec.beam.pulsepot>>htemp","rec.beam.pulsepot>0","goff")
(long long) 11960
root [2] cout<<htemp->GetMean() * htemp->GetEntries()<<endl;
6.07715e+17
root [3] cafTree->GetEntries("rec.nd.minerva.nixn>0 && rec.nd.lar.ndlp>0 ")
(long long) 0
```

We can see that the new beam + non beam file have about 26 more POT and no match Mx2 2x2.

PR #99, opened 2025-05-28
Feature/minerva data processing

noeroy:
Related to issue #97, Several additions:

- Trigger type now is listed to make it less arbitrary and easier to understand in the future. Could be an issue in the link flow io_group -> trigger type, it works or 2x2 but not sure if FSD/NDLar would keep the same link, so I tried to made it clear that it's 2x2 only.
- The trigger matching logics now find the next trigger of same type in the other trigger streams, and then tries to do the matching with those.
- The POT is only fetched for beam trigger.

effect of the fix have been reported page 2 of that presentation:
https://docs.dunescience.org/cgi-bin/sso/RetrieveFile?docid=33729&filename=trigger_issue_v9.pdf&version=1

### DIFF
--- a/src/beam/IFBeam.cxx
+++ b/src/beam/IFBeam.cxx
@@ -56,7 +56,12 @@ namespace cafmaker
       double max_time = std::numeric_limits<double>::lowest();
       for (const auto& group : groupedTriggers) {
           for (const auto& trig : group) {
+              if (trig.second.triggerType !=1) continue;
               double trigger_time = util::getTriggerTime(trig.second);
+
+              // HACK
+              if (trigger_time > 1e12)
+                continue;
               min_time = std::min(min_time, trigger_time - dt);
               max_time = std::max(max_time, trigger_time + dt);
           }
@@ -99,7 +104,9 @@ namespace cafmaker
   
   double IFBeam::getPOT(const cafmaker::Params& par, const TriggerGroup& groupedTrigger, int ii) {
       double pot = 0.0;
-  
+      if (groupedTrigger.front().second.triggerType != 1) return 0.0;
+   //   for (auto trig : groupedTrigger) {std::cout<<trig.second.triggerType<<" ";}
+  //  std::cout<<std::endl;
       auto it = std::find_if(beamSpills.begin(), beamSpills.end(),
           [par, &groupedTrigger](const auto& spill) {
               return std::all_of(groupedTrigger.cbegin(), groupedTrigger.cend(),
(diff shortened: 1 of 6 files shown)


chenel (changes requested):
On the whole this looks fine.  However, I don't really think making `TriggerType` an enum here is the right solution.  If the DAQ for SAND, for example, uses `0` to represent beam triggers instead of `5`, we have no way to do that cleanly with the `enum` (it would require nonsense like
```
enum Trigger Type
{
   kBeamSAND = 0,
   kBeam2x2 = 5,
   ...
}
```
which really doesn't belong in the generic code.)

I think a better (and equally easy) solution would be to define some `int` constants in the ND-LAr-related header that look like
```
const int kBeamTrigger = 5;
const int kLightTrigger = 6;
```
and then simply check `if (trigger.type == kBeamTrigger)` etc. in the branch filler code.  (If you want something more 2x2 specific it's fine to wrap these constants in their own namespace.)  This way the association between specific numbers and the named trigger types is owned by the detector-specific code rather than being put into the general branch filler header.

noeroy:
Alright, couple of updates: 
- I've handled the triggers as suggested (I think) 
 
-  I added a way to handle the trigger type as input properly, with a parameter for the trigger type that is -1 (all triggers) by default. So anything we do today shouldn't be impacted but if  we want to do studies of beam only/non beam only that should be easier. 

- I've made sure that calling CAFMaker with input trigger type = -1 or input trigger type = 1 gives me the same POT, something different than that gives 0, and raises a warning if one of the input file doesn't have any triggers of requested type. 
 
- Sim is not affected by those modifications 

- IFbeam database is queried only if there is at least one beam trigger.

chenel (commented):
This is certainly an improvement, though I think we are not totally there yet.  In particular the hard-coded `1`s in the IFBeam area give me the heebie-jeebies.

If I'm not mistaken--and tell me if I am!---I think there are two considerations we need to handle:

* We need to match triggers across different recos within the same detector
* We need to match _beam_ triggers across detectors (& recos)

I think we *should* be able to assume that the same detector gets the same trigger codes across recos, right?  If so, what I would propose as the right way forward is:

* drop the `triggerMap` thing, which is essentially doubling the set of arbitrary codes that need maintainenance (one of them has to agree with the DAQs, and the other has to be self-consistent across triggerFillers here), and the values that aren't `1` are not used anywhere else?
* add a new field, `isBeam`, to the `Trigger` struct, to be filled by each `RecoBranchFiller` when it fills in the triggers (since each RBF knows which trigger(s) correspond to beam)
* adjust `doTriggersMatch` to also check `trigger.isBeam`
* use `trigger.isBeam` in the `IFBeam` code instead of checking for `1`

What thoughts do you have about this?  (I'm sure there's some edge case I haven't thought of...)

noeroy:
So if I understand, the idea would be that we only make sure that the definition of a "beam trigger" is equivalent in all RBF thanks to a IsBeam function, otherwise when doing matching we can rely on the fact that the trigger type should be consistent for RBF based on the same detector input file?

I see one edge case where for instance TMS and NDLar would store the same "trigger type number" for two different type of things, but that can probably be dealt with more easily than having that map to keep in case for some reason someone decide that for 2x2 calibration runs we use a different io_groups or something. 

Then the other limitation I can think of is if someone wants to load only one specific trigger type, it would only be loading RBF from one detector only (typical use case, loading beam triggers wouldn't get you  Mx2-2x2 matched beam only as we don't use the same trigger type number). That can most likely be addressed in another way using the isBeam function for instance. I don't see another use case where we'd want to only match something accross detector that is not beam.

noeroy:
Yes, I'll do that, thanks for the review!

chenel (approved):
Thanks for working with me on this, I think this final design is a lot better and hides the implementation details inside the detector-specific methods where they should be.  I did notice some weird indentation mismatches (tabs vs spaces?) in a few of the places you changed things---if you could have a quick look to make sure those didn't go haywire that'd be appreciated.  But otherwise, 👍🏻

noeroy:
Pandora's the only BRF that uses tabs instead of two spaces, should I follow that prescription for what I added or just change all the tabs by two spaces? That seems out of the scope of that PR but then I think it's something positive if we want to have some consistency in the future.

PR #101, opened 2025-06-17
Trigger type/time filling updates

mjkramer:
Complements https://github.com/DUNE/ND_CAFMaker/pull/99 (which should be merged first). Linked to https://github.com/DUNE/ndlar_flow/pull/195 and https://github.com/PandoraPFA/LArRecoND/pull/40.

- ~Fix compiler error in TMS filler now that trigger type is an enum~ (no longer necessary)
- ~Set Pandora trigger type properly~ (done in #99)
- Use the new `unix_time_usec` for Pandora trigger times. This makes it possible to match with Mx2 triggers using a < 5ms window.


mjkramer:
Just realized that #99 is undergoing changes which affect the definition of `triggerType`, so reverting this to a draft for now.

noeroy (approved):
Seems good!

I guess SPINE is just updating the old field value?

PR #109, opened 2025-11-03
Feature/ingratta add beam quality info

GianfrancoIngratta:
Some beam-related info to evaluate the beam quality for a given spill is missing in the output CAF file (this was marked a while ago as a TODO in a comment here -> #https://github.com/DUNE/ND_CAFMaker/blob/972f11bc5b69ea1f595e14ed16e09162f512011e/src/beam/IFBeam.cxx )
The current code retrieves only the beam POT from the database, and in this PR I added the horn current, horn polarity, beam position, and intensity info to the output CAF.
The code compiles and has been tested on a few minutes' worth of 2x2 data from last year. See docDB for more details: DUNE-doc-35166-v1


chenel (changes requested):
I don't have any insight into the IF Beam DB itself (hopefully @sindhu-ku can comment on the implementation here), but before we can merge this PR, you'll need to `git rm` all the `.o` files that are included in this PR.  We don't need intermediate build files in the repository 😄

src/beam/IFBeam.cxx
```diff
@@ -91,41 +72,171 @@ namespace cafmaker
           curl_easy_cleanup(curl);
           try {
               auto json_data = json::parse(readBuffer);
+              // std::cout << "readBuffer " << readBuffer << "\n";
```
noeroy: I'd remove the cout instead of comment

src/makeCAF.C
```diff
@@ -444,21 +444,56 @@ void loop(CAF &caf,
...
+        horizontalpos121 = beamManager.getData(par, groupedTriggers[ii], ii, "E:HP121[]");
+        verticalposTGT = beamManager.getData(par, groupedTriggers[ii], ii, "E:VPTGT[]");
+        verticalintTGT = beamManager.getData(par, groupedTriggers[ii], ii, "E:VITGT[]");
+        verticalpos121 = beamManager.getData(par, groupedTriggers[ii], ii, "E:VP121[]");
+        multiwireInfo = beamManager.getData(par, groupedTriggers[ii], ii, "E:MTGTDS[]");
     }
     else
     {
-	pot = par().runInfo().POTPerSpill() * 1e13;
+	potTRTGTD = {par().runInfo().POTPerSpill() * 1e13};
```
noeroy: A tab here?

src/beam/IFBeam.cxx
```diff
@@ -160,6 +271,48 @@ namespace cafmaker
...
+      return getData(par, groupedTrigger, ii, verticalPosTGT);
+
+    else if(deviceName == verticalIntTGTDevice)
+      return getData(par, groupedTrigger, ii, verticalIntTGT);
+
+    else if(deviceName == verticalPos121Device)
+      return getData(par, groupedTrigger, ii, verticalPos121);
+
+    else if(deviceName == multiwireDevice)
+      return getData(par, groupedTrigger, ii, multiwireInfo);
```
noeroy: The two functions and the lists of ifs could be merged into one function and a `std::map<std::sting, BeamInfo> ` that you can initialize in the initialization of the class. I'm happy to have it like it is now but that'd reduce the amount of lines + functions.

src/beam/IFBeam.cxx
```diff
@@ -160,6 +271,48 @@ namespace cafmaker
...
+
+    else
+      return {0.};
```
noeroy: Perhaps add a warning in case we don't find the device case we're requesting (for future updates)

src/beam/IFBeam.cxx
```diff
@@ -43,42 +44,22 @@ namespace cafmaker
           int exponent = std::stoi(match.str(1));
           return std::pow(10, exponent);
       } else {
-          cafmaker::LOG_S("Fetch beam information").WARNING() << "Unknown unit in beam database: " << unit << "\n";
+          // cafmaker::LOG_S("Fetch beam information").WARNING() << "Unknown unit in beam database: " << unit << "\n";
```
noeroy: Do we need to keep that comment?

noeroy (commented):
Nice update, I believe the most important comment is the one to properly link to the new duneanaobj when it will be merged + published in the setup script.

Otherwise, it's a couple of cosmetic comment and a proposal to use a map to link the device name to the beamInfo in the IFBeam.cxx, I'm fine if you decide not to do it though.

src/beam/IFBeam.cxx
GianfrancoIngratta: I'm still thinking about it, and I'm not sure. It might be an unnecessary complication. Maybe the map suits our needs better

chenel (approved):

src/makeCAF.C
```diff
@@ -444,21 +444,56 @@ void loop(CAF &caf,
...
+    caf.sr.beam.potTR101D = (potTR101D.size()==0) ? 0. : potTR101D.at(0);
+    caf.sr.beam.hornI = (hornI.size()==0) ? 0. : hornI.at(0);
+    caf.sr.beam.hornDir = (hornDir.size()==0) ? 0. : hornDir.at(0);
+    caf.sr.beam.horizontalposTGT = horizontalposTGT;
+    caf.sr.beam.horizontalintTGT = horizontalintTGT;
+    caf.sr.beam.horizontalpos121 = horizontalpos121;
+    caf.sr.beam.verticalposTGT = verticalposTGT;
+    caf.sr.beam.verticalintTGT = verticalintTGT;
+    caf.sr.beam.verticalpos121 = verticalpos121;
+    caf.sr.beam.multiwireInfo = multiwireInfo;
```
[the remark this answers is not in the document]
chenel: https://github.com/DUNE/duneanaobj/pull/60 was included in `v3_12_00`, and we're now requiring `v3_14_00`, so we should be good here!

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -533,7 +533,8 @@ namespace cafmaker
         // unix_ts trigger time (seconds)
         trig.triggerTime_s = m_unixTime;
         // unix_time_usec ticks (microseconds) converted to nanoseconds
-        trig.triggerTime_ns = m_unixTimeUsec * 1000;
+        // trig.triggerTime_ns = m_unixTimeUsec * 1000;
```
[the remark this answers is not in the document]
chenel: This is now obsolete, but the PR is doing something different instead.
> trig.triggerTime_ns = m_startTime * 100;

@GianfrancoIngratta so long as you're happy with what it's doing now, we can consider this resolved.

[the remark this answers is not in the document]
GianfrancoIngratta: Hi @chenel , I think we are good on this one. Before closing this, I wanna add a bit of code to address this issue here https://github.com/DUNE/ND_CAFMaker/issues/128. It shouldn't be a big deal...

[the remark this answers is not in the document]
chenel: Okay.  I'll resolve this comment and you can add below.

chenel:
Note to self: Gianfranco is planning to add some more code to this PR to address #128, so don't close yet.

GianfrancoIngratta:
hi @chenel  and @alexbooth92 , here is the commit to address issue #128. To be clear - it compiles, but still needs testing. I didn't have much time yesterday to do so, but I pushed this anyway so we can discuss the proposed solution in the meantime.

alexbooth92 (commented):
Thank you for this implementation, @GianfrancoIngratta. I have only looked at the latest commit regarding MC POT and had to remind myself how all of this works.

In broad strokes, it looks like the loop where you are filling is over spills, so pot per spill is the appropriate value to pull from the edep-sim file. Additionally, you are incrementing `caf.pot` in the same way as for data, so I think all is good.

src/truth/FillTruth.cxx
```diff
@@ -607,6 +619,11 @@ namespace cafmaker
     else {
       fEdepTree=NULL;
     }
+    // Read TParameter<double> pot_per_spill
+    auto* potParam = dynamic_cast<TParameter<double>*>(fEdepFile->Get("pot_per_spill"));
+    if (potParam) {
+      f_POTperSpill = *potParam;
+    }
```
chenel: (note: this comment will become obsolete if the suggestions regarding moving this logic elsewhere are adopted)

if `fEdepFile` winds up `nullptr` (user passed bad filename or whatever), the `fEdepFile->Get()` call is going to segfault.  These lines need to go inside the `if (fEdepFile ...)` block above.

src/truth/FillTruth.cxx
```diff
@@ -607,6 +619,11 @@ namespace cafmaker
     else {
       fEdepTree=NULL;
     }
+    // Read TParameter<double> pot_per_spill
+    auto* potParam = dynamic_cast<TParameter<double>*>(fEdepFile->Get("pot_per_spill"));
+    if (potParam) {
+      f_POTperSpill = *potParam;
```
chenel: (note: this comment will become obsolete if the suggestions regarding moving this logic elsewhere are adopted)

if the key `pot_per_spill` isn't in the file, `f_POTperSpill` is going to remain 0.0, with no warning emitted. This would silently zero out POT for any MC sample that predates this key being written into edepsim files.  I think there probably needs to be a fallback to the configured `POTPerSpill` value and a warning log when it's used.

src/makeCAF.C
```diff
@@ -474,7 +474,8 @@ void loop(CAF &caf,
     }
     else
     {
-	potTRTGTD = {par().runInfo().POTPerSpill() * 1e13};
+	// potTRTGTD = {par().runInfo().POTPerSpill() * 1e13};
+        potTRTGTD = {truthMatcher.GetPOTperSpill().GetVal()};
```
chenel: This drops the `1e13` suffix.  I don't know offhand---is there ever a case where an edepsim file reports in anything other than unscaled POT?

src/makeCAF.C
```diff
@@ -474,7 +474,8 @@ void loop(CAF &caf,
     }
     else
     {
-	potTRTGTD = {par().runInfo().POTPerSpill() * 1e13};
+	// potTRTGTD = {par().runInfo().POTPerSpill() * 1e13};
```
chenel: we have the `git` history.  was there a reason you left this commented out rather than deleting?  if there's something you need to preserve, a better alternative is a `todo:` or `fixme:` comment

src/truth/FillTruth.cxx
```diff
@@ -464,6 +464,18 @@ namespace cafmaker
...
+  {
+    if (HaveEDEPSIM())
+    {
+      return fEdepSimTree.GetPOTperSpill();
+    }
+    else
+    {
+      return TParameter<double>("pot_per_spill", 0.0); 
+    }
+  }
```
chenel: (note: this comment will probably become moot once the others have been addressed, but just in case)

This function should return just a bare `double`, not `TParameter<double>`.  `TParameter` just exists so you can store a value in a ROOT file with a name---but none of that is needed once the value has been read out.

src/truth/FillTruth.h
```diff
@@ -21,6 +21,7 @@
 #include "util/FloatMath.h"
   //TG4Event
 #include "TG4Event.h"
+#include "TParameter.h"
```
chenel: This `#include` won't be necessary once the other comments have been addressed.

src/makeCAF.C
chenel: if you take my overall comment in this review, then this becomes

```
double mcPOTperSpill = readPOTFromEdepsim(edepsimFilename).value_or(par().runInfo().POTPerSpill() * 1e13);
potTRTGTD = {mcPOTperSpill};
```

chenel (changes requested):
I have some comments 😬

First, obviously don't want to merge before you've had a chance to test.  Secondly, I think this approach creates some problems, and I would like it to be handled differently.  I think it would be reasonable to split into another PR, if you don't have time to handle this before merging.

There are a number of nits, but the main concern I have is that POT is file-level beam metadata rather than event-level truth information, so I don't think `TruthMatcher` is the right home for the "true POT" lookup. I'd suggest instead a small free function in `makeCAF.C`, placed between `getRecoFillers()` and `doTriggersMatch()`:

```
std::optional<double> readPOTFromEdepsim(const std::string& filename)
  {
    if (filename.empty()) return std::nullopt;
    std::unique_ptr<TFile> f(TFile::Open(filename.c_str()));
    if (!f || f->IsZombie()) return std::nullopt;
    auto* p = dynamic_cast<TParameter<double>*>(f->Get("pot_per_spill"));
    return p ? std::make_optional(p->GetVal()) : std::nullopt;
  }
```

This would get called once, right before the event loop.  This design also gives a natural place to handle the fallback to the configured value (which is otherwise lost). I'd suggest making `POTPerSpill` in `RunConfig` (i.e., `Params.h`) an `OptionalAtom` so it doesn't need to be specified in fcl files when the edepsim file carries the value. and then the call in loop() becomes:

```
  double mcPOTperSpill = readPOTPerSpillFromEdepsim(edepsimFilename)
                           .value_or(par().runInfo().POTPerSpill() * 1e13);
  potTRTGTD = {mcPOTperSpill};
```

With this approach the `GetPOTperSpill()` methods on TruthMatcher and EdepSimTreeContainer can be removed entirely, and some of the inline comments below become moot.

GianfrancoIngratta:
Hello folks @chenel @alexbooth92 

First, thanks @chenel for providing a loooooot of suggestions 🤣! 
I reverted the last commit, so we are back to whatever was approved before (so I guess we can proceed and merge?)

I'll refer to this conversation in a new PR when I address issue 128 at some point.

chenel:
Sounds good, I think that's the best path forward.  Could you just take a look at the conflicts in the `ndcaf_setup.sh`, which we have to figure out before we can merge?

chenel (approved):
Okay.  I think we're ready to go 🚀

PR #110, opened 2025-11-25
Update Pandora Reco branch filler to use new variables in Pandora output file

GianfrancoIngratta:
In this draft PR we introduce updates to the Pandora Reco branch filler. Recent developments in the post-Pandora workflow (the so-called PandoraOuterface) have added new track/shower-discrimination variables to the Pandora output. The attached image shows the new branches now available in the Pandora output, which we intend to use to populate the SRRecoParticle via the branch filler.

image

### DIFF
--- a/src/reco/PandoraLArRecoNDBranchFiller.cxx
+++ b/src/reco/PandoraLArRecoNDBranchFiller.cxx
@@ -403,7 +403,7 @@ namespace cafmaker
     const float dZ = endZ - startZ;
     const float distStartEnd = sqrt(dX * dX + dY * dY + dZ * dZ);
 
-    // Energy Pandora (MeV) --> CAF (GeV)
+    // Energy Pandora default Hierarchy analysis (MeV) --> CAF (GeV)
     const float energy = (m_energyVect != nullptr) ? (*m_energyVect)[iCluster]/1000. : 0.0;
 
     // Total number of 3D hits in the cluster
@@ -667,9 +667,6 @@ namespace cafmaker
           // Initialise total neutrino energy to zero
           interaction.Enu.calo += track.Evis;
 
-          if (abs(recoParticle.pdg) == abs(m_muonPDG)) 
-            interaction.Enu.lep_calo = longestTrackE + longestTrackCaloE;
-
           sr.nd.lar.pandora[nuIndex].tracks.emplace_back(std::move(track));
           sr.nd.lar.pandora[nuIndex].ntracks++;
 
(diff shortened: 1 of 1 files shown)


chenel (approved):
I don't see any major issues here from the high-level CAFMaker perspective.  I would feel better if @brucehoward-physics  or @jback08 could give their okay that everything checks out before we merge it.  But so long as it doesn't introduce any build issues, I'm okay with going forward if there are other downstream things waiting.

GianfrancoIngratta:
Hi all, just putting some sanity-check plots here. These are all scatterplots with the x-axes being the input variables from the input file (Pandora outerface branch), and on the y-axes the corresponding variables as named in the caf output. If I didn’t inject any weirdness in the code, I should see the same numbers.

I still have to do this check for the other variables, but these plots here are reassuring. 
[sanity_check_plots.pdf](https://github.com/user-attachments/files/24679468/sanity_check_plots.pdf)

GianfrancoIngratta:
In the last commit, I filled the shower attributes introduced in the new duneanaob [PR66](https://github.com/DUNE/duneanaobj/pull/66). 

I have also rearranged the way we fill the SRShower information and its associated
SRRecoParticle. Previously, we filled SRRecoParticle first and then copied
the relevant fields into SRShower; now the flow is reversed. 

I introduced this change while adding new fields to SRShower and realized
That restructuring the filling logic avoids accessing the same shower
variables multiple times from the input tree, simplifying the code and
removing redundant reads.

The same change could be applied to the SRTrack part, but not really necessary if not for aesthetic reasons.

I've run the checksum on the new output file, checking that I have not introduced bugs when introducing these changes.

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -28,6 +30,11 @@ namespace cafmaker
         throw;
       }
 
+      // Check if input file has Pandora Outerface fields. If false, fill
+      // Standard Record classes with default filler
+      const bool inputHasOuterfaceBranches = HasOuterfaceBranches();
+      std::cout << "Input file has Pandora Outerface branches? " << inputHasOuterfaceBranches << "\n";
```
jback08: Change cout to LOG.VERBOSE()

src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -428,64 +558,170 @@ namespace cafmaker
...
+      const int trackIxn = trackTrueID.ixn;
+      const float trackOverlap = (truthOverlap.size() > 0) ? truthOverlap[0] : 0.0;
 
       // Is reco primary & reco PDG hypothesis
       const int isRecoPrimary = (m_isRecoPrimaryVect != nullptr) ? (*m_isRecoPrimaryVect)[i] : 0;
-      showerPart.primary = (isRecoPrimary == 1) ? true : false;
-      showerPart.pdg = (m_recoPDGVect != nullptr) ? (*m_recoPDGVect)[i] : 0;
+      recoParticle.primary = (isRecoPrimary == 1) ? true : false;
+
+      if(m_trackScoreVect != nullptr)
```
jback08: Add space between if and the bracket, as well as on the next if statement.

PR #123, opened 2026-02-27
Migrate from make to cmake and add GitHub Actions CI workflow

### LINKED ISSUE #11: build is very fragile

### LINKED ISSUE #47: Continuous integration
It would be spectacular if we could get some continuous integration set up in order to test that PRs build and work before we merge them.  We need a hero...

slanzi00:
Resolves #11 and #123

**#11 — CMake migration:** Replace the fragile GNU Make build with a proper CMake
setup. Non-CMake-aware UPS packages are wrapped as `INTERFACE IMPORTED` targets
(`deps::hdf5`, `deps::genie`, etc.); CMake-aware packages use the standard
`find_package(...)`. Build artifacts are isolated under `build/`,
the project version is pinned to 5.1.0, optional features are exposed as CMake
options (`ENABLE_TMS`, `ENABLE_TESTEXE`), SANDReco is auto-detected from environment
variables, and the compiler warning set has been expanded.

`ndcaf_setup.sh` is split into two scripts: `ndcaf_setup_deps.sh`, sourced before
compilation to set up all required UPS dependencies, and `ndcaf_setup.sh`, which
internally sources the former and is meant to be sourced after installation to
configure the runtime environment.

**#123 — CI & fixes:** Integrate the CMake build from `11-build-is-very-fragile`,
add `.github/workflows/ci.yml` to run on every PR using Apptainer + CVMFS inside
the `fnal-wn-sl7` SL7 container, and fix a compilation error in
`SANDRecoBranchFiller.cxx` where the no-SAND stub for `GetTriggers` was missing
the `bool beamOnly` parameter, causing a signature mismatch with the header.


CMakeLists.txt
```diff
@@ -0,0 +1,218 @@
...
+add_library(deps::root_glibs INTERFACE IMPORTED)
+set_target_properties(
+  deps::root_glibs
+  PROPERTIES
+    INTERFACE_LINK_LIBRARIES
+    "ROOT::Core;ROOT::MathMore;ROOT::Geom;ROOT::EGPythia6;ROOT::GenVector")
+
+# Versions are pinned by ndcaf_setup.sh; env vars are set by UPS at runtime.
+
+find_program(GENIE_CONFIG genie-config HINTS "$ENV{GENIE_INC}/../bin"
```
chenel: usually in GENIE configurations, the top-level variable is `$GENIE`.  I think that's probably worth using instead of `$GENIE_INC/..`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
+name: CI
+
+on:
+  push:
```
chenel: I think we probably don't need to build _every_ time somebody pushes, do we?  Probably it's better to only run the workflow when a PR is made?   And maybe set up a keywords-based trigger so that folks can manually request a build if they want one.

chenel (commented):
This is brilliant, thanks so much for implementing it.  Bravo!  👏🏻 

This will be a large-scale breaking change for a number of workflows, so we'll need to make sure we shop it around before merging it.  Stakeholders I can think of:
* 2x2 production
* ND Reco/Sim production
* Expert ND_CAFMaker users from ND-LAr/2x2, TMS, SAND

We should start by discussing this at a ND CAFMaker scrum on a Wednesday and discuss the plan there.  So I won't approve it yet, not because I think there's anything wrong with it necessarily, but just to make sure nobody accidentally clicks "merge" before we are ready.

Once the CI workflow part of this PR works, I imagine making it a requirement that CI pass before a PR can be merged, which would be a dream come true.  (Even more than that, I imagine adding a _testing_ CI workflow to ensure that future changes don't break the output except when we intend to.  Etc.  But obviously that's not for this PR.)

CMakeLists.txt
```diff
@@ -0,0 +1,218 @@
...
+    OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
+    RESULT_VARIABLE _git_rc)
+  if(_git_rc EQUAL 0)
+    string(REGEX REPLACE "^v" "" _version "${_git_tag}")
+  else()
+    set(_version "0.0.0")
+  endif()
+else()
+  set(_version "0.0.0")
+endif()
```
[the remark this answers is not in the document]
alexbooth92: We have an ["install" script](https://github.com/DUNE/ND_Production/blob/main/run-cafmaker/install_cafmaker.sh) that is run once on a disk space that the grid nodes have access to. It does a clone of `ND_CAFMaker` and then runs `build_deps.sh`.

alexbooth92:
> This will be a large-scale breaking change for a number of workflows, so we'll need to make sure we shop it around before merging it. Stakeholders I can think of:
> 
> * 2x2 production
> * ND Reco/Sim production
> * Expert ND_CAFMaker users from ND-LAr/2x2, TMS, SAND

From the ND Prod side, I've made a few changes to the run and install scripts in this (currently draft) [PR](https://github.com/DUNE/ND_Production/pull/77).

alexbooth92 (approved):

PR #131, opened 2026-04-05
Unique match reco filler check cheated vectors nonempty

QWeyrich:
Added check that the truth overlap and truth id arrays for a LAr track are nonempty (and equal in length) before attempting to carry out the cheated matching


chenel (approved):
small nits, but good

PR #133, opened 2026-04-27
129 fill new reco object parent and corresponding reco particle fields

mt82:
### Summary
Implements bidirectional linkage between low-level reco objects (tracks/showers) and the corresponding `SRRecoParticle` in the `StandardRecord`, for both SPINE/DLP and Pandora. 

---

### Changes

**`MLNDLArRecoBranchFiller.h/.cxx`**
- Implemented new inner class `MLNDLArRecoParticleMapper`, which encapsulates the SPINE particle ID → `(ixn_idx, prt_idx)` mapping with a safe interface:
  - `GetRecoParticleID(partID)` — returns the corresponding `SRRecoParticleID`
  - `GetRecoParticle(sr, partID)` — returns a reference to the `SRRecoParticle`
  - `Reset()` — clears the map at the start of each event, before reading HDF5 products
- `track.part` and `shower.part` are now filled via `GetRecoParticleID(part.id)`
- `srPart.recoobj` is now filled for both tracks and showers with the correct `ixn_idx`, `robj_idx`, and collection type

**`PandoraLArRecoNDBranchFiller.cxx`**
- `SRRecoParticleID` and `SRRecoBaseID` are now constructed and filled in both `FillClusterDefault` and the main particle loop
- `track.part`, `shower.part`, and `recoParticle.recoobj` are set before the objects are pushed into their respective vectors

---

### Notes
- The SPINE/DLP implementation has been tested on real input files. The Pandora implementation compiles but could not be fully tested due to missing input files

### DIFF
--- a/src/reco/MLNDLArRecoBranchFiller.cxx
+++ b/src/reco/MLNDLArRecoBranchFiller.cxx
@@ -117,6 +117,23 @@ namespace cafmaker
 
   }
 
+  // ------------------------------------------------------------------------------
+  // helper class to map from SPINE track ID to (sr.common.ixn.dlp, sr.common.ixn.dlp.part.dlp) indices for the corresponding SRRecoParticle
+  caf::SRRecoParticleID MLNDLArRecoBranchFiller::MLNDLArRecoParticleMapper::GetRecoParticleID(int64_t partID) const
+  {
+    if(fParticleMap.find(partID) == fParticleMap.end())
+    {
+      LOG.FATAL() << "MLNDLArRecoParticleMapper: could not find particle ID " << partID << " in the particle map! Abort.\n";
+      abort();
+    }
+    auto [ixn_idx, prt_idx] = fParticleMap.at(partID);
+    return caf::SRRecoParticleID{ixn_idx, caf::SRRecoParticleID::SRRecoParticleCollectionType::kSPINE, prt_idx};
+  }
+
+  caf::SRRecoParticle& MLNDLArRecoBranchFiller::MLNDLArRecoParticleMapper::GetRecoParticle(caf::StandardRecord & sr, size_t ixn_idx, size_t prt_idx) const
+  {
+    return sr.common.ixn.dlp[ixn_idx].part.dlp[prt_idx];
+  }
 
   // ------------------------------------------------------------------------------
   // todo: possibly build some mechanism for customizing the dataset names in the file here
@@ -173,7 +190,7 @@ namespace cafmaker
     sr.meta.lar2x2.readoutstart_ns = trigger.triggerTime_ns;
     
     // reset the map of SPINE particle ID to (sr.common.ixn.dlp, sr.common.ixn.dlp.part.dlp) indices for this entry
-    fParticleMap.clear();
+    fParticleMapper.Reset();
 
     H5DataView<cafmaker::types::dlp::Interaction> interactions = fDSReader.GetProducts<cafmaker::types::dlp::Interaction>(idx);
     H5DataView<cafmaker::types::dlp::TrueInteraction> trueInteractions = fDSReader.GetProducts<cafmaker::types::dlp::TrueInteraction>(idx);
@@ -579,7 +596,7 @@ namespace cafmaker
       // index of the particle within the sr.common.ixn.dlp.part.dlp vector
       auto prt_idx = sr.common.ixn.dlp[ixn_idx].part.dlp.size();
       // save the indices for this particle so we can get it back later
-      fParticleMap[part.id] = std::make_pair(ixn_idx, prt_idx);
+      fParticleMapper[part.id] = {ixn_idx, prt_idx};
       // fill the reco particle
       sr.common.ixn.dlp[ixn_idx].part.dlp.push_back(std::move(reco_particle));
       sr.common.ixn.dlp[ixn_idx].part.ndlp++;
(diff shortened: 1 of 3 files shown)


src/reco/PandoraLArRecoNDBranchFiller.cxx
```diff
@@ -755,6 +784,9 @@ namespace cafmaker
         LOG.DEBUG() << "trackfit failed for particle number : " << i << ", assigning pdg 0\n";
         FillClusterDefault(sr, i, uniqueSliceIDs, nuInteractions, truthMatch, longestTrack, longestTrackDir, maxShowerE, maxShowerEDir);
       }
+      
+      // Set the recoobj ID for the particle
+      recoParticle.recoobj = recoBaseID;
```
chenel: If the track fit failed, I think `recoBaseID` is left unset.  So `recoobj` is being created with default-constructed values (==nothing).  This seems like it should never happen, unless I'm missing something?  It might be worth warning (or aborting?) in the `else` clause above?

src/reco/MLNDLArRecoBranchFiller.cxx
```diff
@@ -117,6 +117,40 @@ namespace cafmaker
...
+    if(fParticleMap.find(partID) == fParticleMap.end())
+    {
+      LOG.FATAL() << "MLNDLArRecoParticleMapper: could not find particle ID " << partID << " in the particle map! Abort.\n";
+      abort();
+    }
+    auto [ixn_idx, prt_idx] = fParticleMap.at(partID);
+    return caf::SRRecoParticleID{static_cast<int>(ixn_idx), caf::SRRecoParticleID::SRRecoParticleCollectionType::kSPINE, static_cast<int>(prt_idx)};
+  }
+
+  caf::SRRecoParticle& MLNDLArRecoBranchFiller::MLNDLArRecoParticleMapper::GetRecoParticle(caf::StandardRecord & sr, int64_t partID) const
```
chenel, line 135: Claude made a good suggestion here: the fact that the object returned is not `const` but the method is is potentially confusing.  A comment like the following would be helpful:

```suggestion
  // The method is `const` because it applies to the _mapper_--- we're not changing the mapping---
  //  but the returned instance is part of the SR itself and can be modified
  caf::SRRecoParticle& MLNDLArRecoBranchFiller::MLNDLArRecoParticleMapper::GetRecoParticle(caf::StandardRecord & sr, int64_t partID) const
```

chenel (approved):
Thank you for doing this!!  Overall things look great.  Detailed comments below.

One thing for thinking about: I asked Claude to review this PR also in order to catch anything I didn't see (since it's a fairly extensive changeset), and one thing it pointed out is that there's an implicit interdependency in the order between creation of particles and the "underlying" objects.  It might be prudent for us to think up some "locking" mechanism that guarantees that either all the `SRRecoParticle`s are created before all the `SRRecoBase`-derived objects, or vice versa.  Would appreciate your input on that. 
(One possible way of implementing that would be a simple `fParticlesFilled` that is initialized to `false` and is set to `true` after they're filled in.)

mt82:
That’s a good observation, and I agree it’s worth thinking about.

In `MLNDLArRecoBranchFiller.cxx`, all `SRRecoParticles` are created first, and only afterwards are the underlying `SRRecoBase`‑derived objects filled. The current implementation of the link between them implicitly assumes this specific ordering, since it is effectively enforced by the way the code is structured today.

In `PandoraLArRecoNDBranchFiller.cxx`, however, the situation is different: for each particle reconstructed by Pandora, the two objects (`SRRecoParticle` and the corresponding `SRRecoBase`‑derived object) are created almost simultaneously.

In my opinion, ensuring a robust and well‑defined algorithm to implement the link between particles and reco‑objects should be the responsibility of the concrete `RecoBranchFiller` implementation itself. 

If I understood correctly, you are proposing to add an `fParticlesFilled` flag to `IRecoBranchFiller.h`, which would be set to `true` once all `SRRecoParticles` have been created. Introducing a simple locking or state mechanism based on this flag could certainly help in the case of `MLNDLArRecoBranchFiller.cxx`, where the particle creation phase is clearly separated. However, it would likely not work as-is for `PandoraLArRecoNDBranchFiller.cxx`, where particle and reco-object creation are inherently interleaved.

chenel:
Okay---I don't think I appreciated the difference in the pathways through the filling.  In that case, I think we probably can't enforce something that's safer, really, so maybe we should just press on as is.

mt82:
## Validation Scripts and Checks

The scripts used for these checks are available here: [scripts.tar.gz](https://github.com/user-attachments/files/28286595/scripts.tar.gz)

Please unpack the archive on a `dunegpvm` machine and run:

```bash
./run-all.sh
```

---

## 1st Check — Reconstructed Energy Comparison

An initial validation was performed by comparing reconstructed energy for matched categories of objects in CAF and SPINE, using the following selections:

- **CAF**
  - `SRTrack` (`SRShower`)
  - Associated `SRRecoParticle` (via `FindRecoParticle`)
  - With `contained == true` (`false`)

- **SPINE**
  - Particle with `shape == kTrack` (`kShower`)
  - With `is_contained == true` (`false`)

The comparison results are shown below:

#### Uncontained Track
diff_trk_ucntd_mcs_ke_E

#### Contained Track
diff_trk_cntd_csda_ke_E

#### Uncontained Shower
diff_shw_ucntd_calo_ke_E

#### Contained Shower
diff_shw_cntd_calo_ke_E

After running the scripts, the output can be found in the `ana` directory, including:

- `reco_particles.root`
- All generated `.png` comparison plots

---

## 2nd Check — `obj.len_cm` Consistency

For each `SRRecoObjBase& obj` inside every `SRInteraction` within each `SRStandardRecord`, we verified the consistency of:

```cpp
obj.len_cm == obj.part->recoobj.len_cm
```

PR #141, opened 2026-05-06
140 bug duplicate srrecoparticle creation when m trackscorevect is nullptr in fillrecoparticles

### LINKED ISSUE #140: Bug: Duplicate SRRecoParticle creation when m_trackScoreVect is nullptr in FillRecoParticles
## Summary
When `PandoraLArRecoNDBranchFiller::FillRecoParticles` is run with `m_trackScoreVect == nullptr`, two `caf::SRRecoParticle` objects are created and added per cluster. This appears to be unintended or likely a bug.

## Details
- In the block where `m_trackScoreVect == nullptr` (i.e., track fit information is missing), the code enters the `else` path and calls `FillClusterDefault()`, which:
  - Creates and adds one `SRRecoParticle` to `interaction.part.pandora`.
- After returning from `FillClusterDefault`, the main loop in `FillRecoParticles` continues and another (likely uninitialized) `SRRecoParticle` is created and added to `interaction.part.pandora`.

## Consequences
This means that, in events without track fit information, two `SRRecoParticle` objects are created for each cluster, while only one is expected. This likely leads to duplicate or incomplete particles being stored in the CAF output, potentially affecting downstream analyses.

mt82:
## Summary
Fixes a bug in `PandoraLArRecoNDBranchFiller.cxx` where `SRRecoParticle` objects were being created and added to the interaction even when `m_trackScoreVect` is `nullptr`, causing duplicate and invalid particles to be added to the CAF file.

## Changes
- **Moved particle creation inside the null-check block**: The `SRRecoParticle` instantiation and initialization (lines 617-625) are now only performed when `m_trackScoreVect != nullptr`, preventing creation of incomplete particles
- **Moved particle addition inside the null-check**: The code that appends the particle to `interaction.part.pandora` and adds truth info to the interaction (lines 753-759) is now only executed when trackfit succeeds

## Impact
This prevents duplicate `SRRecoParticle` entries in the output when trackfit fails or when `m_trackScoreVect` is unavailable, ensuring data integrity in the CAF file.

### DIFF
--- a/src/reco/PandoraLArRecoNDBranchFiller.cxx
+++ b/src/reco/PandoraLArRecoNDBranchFiller.cxx
@@ -310,7 +310,7 @@ namespace cafmaker
      shower.direction = dir;
      shower.time = -999.; // TODO to be filled at some point
      shower.Evis = (*m_shwrEnergy)[i]/1000.; // [GeV]
-     shower.qual = (*m_trackScoreVect)[i]; // saving the trackScore value as additional reco info. This provides a sort of degree of “shower-likeness” for this SRShower
+     shower.qual = (*m_trackScoreVect)[i]; // saving the trackScore value as additional reco info. This provides a sort of degree of "shower-likeness" for this SRShower
      shower.len_cm = shwrLength;
      shower.initial_dEdx = dEdx;
      shower.conversionGap = conversionGap;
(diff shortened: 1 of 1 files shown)


GianfrancoIngratta:
Hi @mt82! Thanks for spotting this issue! I think the fix is right -- when the condition `(m_trackScoreVect != nullptr)` [here](https://github.com/DUNE/ND_CAFMaker/blob/bcb10db3c7df6e43d6b6b2439dee7895e569d71f/src/reco/PandoraLArRecoNDBranchFiller.cxx#L622) isn't satisfied, we still want to fill the reco particle with some sort of default information which effectively we do when calling `FillClusterDefault(...)`  in line [765](https://github.com/DUNE/ND_CAFMaker/blob/bcb10db3c7df6e43d6b6b2439dee7895e569d71f/src/reco/PandoraLArRecoNDBranchFiller.cxx#L765). Good catch!

GianfrancoIngratta (approved):

PR #147, opened 2026-05-17
Fix TMS reco truth particle IDs so FindParticle() returns valid pointers

### LINKED ISSUE #146: Retrieving TMS particle objects from their true IDs gives null pointer
The `truth` array of the tracks coming from TMSReco is supposed to store the particle IDs of the true tracks that contributed hits to the track, such that calling `FindParticle(sr.mc,partID)` using a `partID` of a specific particle in the `truth` array should produce the original Monte-Carlo particle object from Geant4. However, this seems to consistently output a null pointer instead.

Image

Image

chenel:
Thanks to Claude I can offer you a suggested fix for this quickly.  @qweyrich please read it carefully, as I'm not a TMSReco expert, but I did vet what came out of Claude and it looks like it's probably right.

--------

Fixes #146.

`TrueParticleID` fields `ixn` and `part` were both set incorrectly when filling truth information for TMS reco tracks:

- `ixn` was set to the raw GENIE interaction ID (run×10⁶ + vertex) rather than the vector index into `sr.mc.nu`, which is what `FindParticle()` uses as an array subscript.
- `part` was never set, leaving it at its default of -1. `FindParticle()` returns `nullptr` immediately when `part < 0`, which is why the lookup always failed.

Both fields are now computed using `std::distance` lookups, following the same pattern as `MLNDLArRecoBranchFiller`. The primary/secondary distinction for TMS tracks remains a TODO (see commented-out `_RecoTruePartIdSec` note in the code).

### DIFF
--- a/src/reco/TMSRecoBranchFiller.cxx
+++ b/src/reco/TMSRecoBranchFiller.cxx
@@ -135,9 +135,6 @@ namespace cafmaker
     LastSpillNo = _SpillNo;
 
     caf::SRTMSInt *interaction;
-    caf::TrueParticleID truePartID;
-    caf::SRTrueParticle srTruePart;
-    caf::SRTrueInteraction  srTrueInt;
 
     // Fill Truth parts first?
     for (int i_tru=0; i_tru< TMSTrueSpill->GetEntries(); i_tru++)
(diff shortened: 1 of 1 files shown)


QWeyrich (approved):
From what I can tell, it works as one would expect. I fed it one Pandora file and one corresponding TMSReco file (`/exp/dune/app/users/weyrich/CAFMaker_ingratta/ND_CAFMaker/cfg/QWeyrich/pandora_outerface_N4p1_pCheat_leadShow_quinton_time.fcl`) and it found 27 true matches by particle ID, and no null pointers. I'm approving this change. Glad to have the TMS particle IDs working, as it will make verifying the performance of the matcher far easier!

Image

QWeyrich:
Note: In order to test this I needed to checkout an older version of `main` (`c83b6267e2ed11f2d454c2eeecccb23cfe3a417a`) since the newer versions have an error when TMS writes the CAF which is preventing ND CAFMaker from running. This is a known issue and is being looked into

Issue #116, opened 2026-02-13
Linkage of Primary and Secondary Particles


alexbooth92:

jdkio:
I have some experience with edep sim particle trajectories. I'd need to familiarize myself with how it's all implemented here so it's not a trivial on-ramp for me, but I'm happy to help out

jdkio:
I've fixed in this PR #150. Who should I assign to review? @chenel?

jdkio:
PR #150 fixes this issue. Closing

PR #150, opened 2026-06-02
Fixes the connection between primary and secondary particles

jdkio:
Properly connects secondary particles to the parent particles in the caf, see issue #116. Also adds validation script.

```bash
# Before fix
=== validation summary ===
file: default/example-CAF.root
entries checked: 13
secondaries checked: 10426
ancestor refs resolved: 0
ancestor chain matches where CAF chain is complete: 0
errors: 10426
warnings: 0
FAIL

# After fix:
python scripts/validate_caf_truth.py fixed_secondaries/example-CAF.root 
=== validation summary ===
file: fixed_secondaries/example-CAF.root
entries checked: 13
secondaries checked: 10426
ancestor refs resolved: 10426
ancestor chain matches where CAF chain is complete: 10426
errors: 0
warnings: 0
PASS
```

The file sizes are not that different
```bash
3.5M    default/example-CAF.root
3.5M    fixed_secondaries/example-CAF.root
```
Run times are not either (note these were measured on two different days):
```bash
Original: 3m31s
Fixed: 3m29s
```

# Note about intermediate secondaries
Secondaries are only saved if a reconstructed track warrants it. This makes sense so we're not save too much info. But this meant that some intermediate particles weren't saved. Like if a secondary neutron creates a proton track, which we reconstruct, then there was no parent to the proton in the caf. Below is the result of the validation if we don't save the interemediate particles. The fix was to save the intermediate particles. It's a more general and stable solution than trying to link the secondary to the primary. Currently the unreco'd particles are not individually flagged but we could add that if need be.

```bash
python scripts/validate_caf_truth.py particle_fix/example-CAF.root 
=== validation summary ===
file: particle_fix/example-CAF.root
entries checked: 13
secondaries checked: 7680
ancestor refs resolved: 7680
ancestor chain matches where CAF chain is complete: 5279
ancestor refs not independently checkable because CAF parent is missing: 2401
trace_missing-parent: 2401
errors: 0
warnings: 0
PASS
```

# Files touched
`src/truth/FillTruth.cxx`
`.gitignore`

# Files added
`scripts/validate_caf_truth.py`

### DIFF
--- a/src/truth/FillTruth.cxx
+++ b/src/truth/FillTruth.cxx
@@ -531,6 +531,76 @@ namespace cafmaker
 
       return ancestor;
     }
+
+    bool HasParticleWithG4ID(const std::vector<caf::SRTrueParticle> &collection,
+                             int G4ID)
+    {
+      return std::find_if(collection.begin(), collection.end(),
+                          [G4ID](const caf::SRTrueParticle &part)
+                          {
+                            return part.G4ID == G4ID;
+                          }) != collection.end();
+    }
+
+    void FillParticleFields(caf::SRTrueParticle &part,
+                            const caf::SRTrueInteraction &ixn,
+                            std::size_t nixn,
+                            int G4ID,
+                            const TG4Event *g4event)
+    {
+      auto traj = g4event->Trajectories[G4ID];
+
+      part.G4ID = traj.TrackId;
+      part.interaction_id = ixn.id;
+      part.pdg = traj.PDGCode;
+      part.p = traj.InitialMomentum*0.001;
+
+      part.parent = traj.ParentId;
+      part.ancestor_id = FindPrimaryAncestor(ixn, nixn, G4ID, g4event);
+
+      auto p0 = traj.Points[0];
+      part.start_pos = (p0.Position * .1).Vect();
+
+      auto pf = traj.Points[traj.Points.size()-1];
+      part.end_pos = (pf.Position * .1).Vect();
+    }
+
+    void EnsureSecondaryParentClosure(caf::SRTrueInteraction &ixn,
+                                      std::size_t nixn,
+                                      int G4ID,
+                                      std::vector<caf::SRTrueParticle> &secondaries,
+                                      int &counter,
+                                      const TG4Event *g4event)
+    {
+      if (!g4event) return;
+      if (G4ID < 0 || G4ID >= static_cast<int>(g4event->Trajectories.size())) return;
+
+      int current = g4event->Trajectories[G4ID].ParentId;
+      std::vector<int> visited{G4ID};
+
+      while (current >= 0)
+      {
+        if (std::find(visited.begin(), visited.end(), current) != visited.end())
+          return;
+        visited.push_back(current);
+
+        if (current >= static_cast<int>(g4event->Trajectories.size()))
+          return;
+
+        if (HasParticleWithG4ID(ixn.prim, current))
+          return;
+
+        if (HasParticleWithG4ID(secondaries, current))
+          return;
+
+        const int particle_index = counter;
+        secondaries.emplace_back();
+        counter++;
+
+        FillParticleFields(secondaries.at(particle_index), ixn, nixn, current, g4event);
+        current = g4event->Trajectories[current].ParentId;
+      }
+    }
   }
 
   int TruthMatcher::FillParticle(caf::SRTrueInteraction &ixn, std::size_t nixn, int G4ID, std::vector<caf::SRTrueParticle> & collection, int & counter, const TG4Event * g4event)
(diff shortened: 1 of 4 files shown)


src/truth/FillTruth.cxx
```diff
@@ -486,54 +486,136 @@ namespace cafmaker
...
+                          }) != collection.end();
+    }
+
+    void FillParticleFields(caf::SRTrueParticle &part,
+                            const caf::SRTrueInteraction &ixn,
+                            std::size_t nixn,
+                            int G4ID,
+                            const TG4Event *g4event)
+    {
+      auto traj = g4event->Trajectories[G4ID];
```
chenel: `auto traj` is copying a full `TG4Trajectory` object here.  I think this could be `const auto & traj = ...`, which won't waste the resources of a copy, and will otherwise function the same below

src/truth/FillTruth.cxx
```diff
@@ -486,54 +486,136 @@ namespace cafmaker
...
+
+      part.G4ID = traj.TrackId;
+      part.interaction_id = ixn.id;
+      part.pdg = traj.PDGCode;
+      part.p = traj.InitialMomentum*0.001;
+
+      part.parent = traj.ParentId;
+      part.ancestor_id = FindPrimaryAncestor(ixn, nixn, G4ID, g4event);
+
+      auto p0 = traj.Points[0];
```
chenel: similar `const auto &` comment here

src/truth/FillTruth.cxx
```diff
@@ -486,54 +486,136 @@ namespace cafmaker
...
+      part.start_pos = (p0.Position * .1).Vect();
+
+      auto pf = traj.Points[traj.Points.size()-1];
```
chenel: ... and here.  also, this could perhaps be more readibly `traj.Points.back()`

src/truth/FillTruth.cxx
```diff
@@ -486,54 +486,136 @@ namespace cafmaker
...
-        ancestor_id = ((*itPart)).ancestor_id.part;
-        // ancestor_id;
+        if (std::find(visited.begin(), visited.end(), current) != visited.end())
+          return;
+        visited.push_back(current);
+
+        if (current >= static_cast<int>(g4event->Trajectories.size()))
+          return;
+
+        if (HasParticleWithG4ID(ixn.prim, current))
```
chenel: Claude noticed the following potential failure mode:

Here you use ixn.prim as a proxy for "is this  a G4 primary?", but `ixn.prim` only contains _reconstructed_ primaries. The reliable signal  is in the G4 trajectories themselves: a true primary has ParentId < 0.

The fix is one extra early-exit check inside the loop, before doing the emplace_back:

```
// current is a G4-level primary (no parent) — stop, but don't add it to sec
if (g4event->Trajectories[current].ParentId < 0)
return;
```

This goes after the bounds check and after the HasParticleWithG4ID checks. The walk stops cleanly at the last intermediate secondary without incorrectly adding the primary to
ixn.sec. The secondary's ancestor_id will remain kUnknown in this case (since the primary isn't in ixn.prim), which is the honest answer.

Assuming that there are no edge cases where a particle in `ixn.prim` doesn't have `ParentId < 0`, this solution would also make the `HasParticleWithG4ID(ixn.prim, ...)` check redundant.

chenel (approved):
This is a great implementation, and I'm especially happy to see the validation script---that'll be a big win in the future if/when we have to check whether future changes break anything.  I'm happy to approve.

There are a few nits recorded below, with one possible edge-case bug that I'd like your further opinion on.

src/truth/FillTruth.cxx
jdkio: This is an interesting edge case. I definitely agree with adding the check and I'm implementing it. The only question is what to do when a primary isn't in the primary list because it wasn't reconstructed. Right now it was accidentally added to the secondaries list. That's not perfect but at least the information is available to the end users. With this new fix, the particle doesn't get added. That could hurt some studies

To me, the fix is to add the missing particle to `prim`, like we do with secondaries that are intermediate parents to particles we reconstruct. Should I implement that? Or do you want `prim` to be exclusive to reconstructed primary particles?

chenel: Yeah, I think adding it to `prim` would be the right move.  I think it might be worth emitting a warning if/when this happens, because I don't know what the frequency is like.  And if you feel particularly motivated to add a counter that emits a summary at job end, that would be bonus points, but certainly not necessary.

jdkio: I added code to add them to prim but my stats checking didn't find any. Looking into it, I'm pretty sure all primary particles are added to the list at the start. So it's not only reconstructed primary particles. So the code should be fine as is

My test file gets these stats (use `export ND_CAFMAKER_STATS=1` to enable):
`TruthMatcher materialization stats: prim=0, sec(requested)=7237, sec(parent-closure)=3189, sec(total)=10426; closure calls=7237, resolved(existing prim)=5306, resolved(existing sec)=1931, resolved(materialized prim)=0, aborted(cycle)=0, aborted(out-of-range)=0`

chenel:
Okay---sounds good.  In that case, if you have the check already instrumented, it might even be worth contemplating whether we should make it an abort, so we know to go figure out what happened?

jdkio:
I'm just going to merge what we have. It'll print a warning. If we want to abort, then we should do a large scale test to make sure it never happens first. My tests weren't extensive enough to be sure

Issue #132, opened 2026-04-17
Spike in true energy distribution


jacoblarkin:
MicroProdN4p1 CAF respin (done with v4.11.1) has strange spikes in the true energy distribution. When looking at 5 random files from this dataset, the spikes in the energy distribution appeared at different locations in different files and some had a single spike while others had multiple. This behavior was not seen in the original MicroProdN4p1 CAFs, made with v4.8.1.

Example plots shown in ND sim/reco meeting on April 10, 2026.

alexbooth92:
Adding a link to the meeting on April 10th https://indico.fnal.gov/event/74007/. See slide 5 of Jacob's talk.

mt82:
We investigated the origin of the spikes observed in the reconstructed neutrino energy distribution and noticed that they are strongly correlated with similar structures in the distribution of the interaction vertex Z coordinate. This correlation is clearly visible in the 2D plot, where the x‑axis represents the Z coordinate of the interaction vertex and the y‑axis represents the neutrino energy. 

<img width="696" height="472" alt="Image" src="https://github.com/user-attachments/assets/e6c38e12-c3dc-4d3e-859e-677077770875" />

This behavior suggests that a subset of events is being duplicated. To validate this hypothesis, we inspected the content of the CAF trees using
```C++
cafTree->Scan("rec.mc.nu.E:rec.mc.nu.id", "", "", 1000, 0);
```
(see the attached file [nu_E_vs_nu_id.txt](https://github.com/user-attachments/files/27201621/nu_E_vs_nu_id.txt)). In the first spill there are 893 neutrino interactions in total, and starting from entry 534 we observe repeated events with exactly the same neutrino energy. This points to the same Monte Carlo neutrino being copied multiple times within the same spill.

Further evidence comes from the comparison of the number of interactions per spill with respect to the previous production. In the older production, the number of interactions per spill is narrowly distributed around an average value of about 93, while in the current production the average increases to 880 interactions per spill. 

<img width="696" height="472" alt="Image" src="https://github.com/user-attachments/assets/8d27262c-361d-4f7c-a12e-eabce0e07bbd" />

<img width="696" height="472" alt="Image" src="https://github.com/user-attachments/assets/c4473919-1fff-4c74-8244-36a488c925f8" />

To further investigate this issue, we attempted to reproduce the CAFs using ND_CAFMaker v4.11.1 with the this fcl configuration. 
```
#include "NDCAFMaker.fcl"
[...]

PR #151, opened 2026-06-03
132 Fix SelectEvent by Correctly Mapping Event Num to Entry Num

jdkio:
Fixes the SelectEvent function. The issue is from a mismatch between event ID variables and entry numbers. Rock files only save events that have neutrinos of high enough energy to hit the rock box, but the event ids are indexed as if all entries are saved

The solution was to create a map on opening the files. Then that map can be quickly referenced rather than scanning per SelectEvent call. Specifically, it's a map of maps to create unique mapping from (run, event num) -> (run, entry num). ...Map

There are a bunch of checks to make sure every mapping is unique and every SelectEvent found the actual run and entry. It will fatally crash if this is not the case, so this should fix the duplicate data issue in #132

# Changed files
src/truth/FillTruth.cxx
src/truth/FillTruth.h

### DIFF
--- a/src/reco/TMSRecoBranchFiller.cxx
+++ b/src/reco/TMSRecoBranchFiller.cxx
@@ -73,17 +73,21 @@ namespace cafmaker
       //TMSRecoTree->SetBranchAddress("TimeSliceStartTime",    &_TimeSliceStartTime);
       TMSLCTree->SetBranchAddress("TMSStartTime",            _TMSStartTime); // Temporary for prod n4p1, time avialable in Reco_Tree for future prods
 
-      // Add Truth tree for the index of the true primary particles
-      TMSTrueTree->SetBranchAddress("TrueVtxX",                       _TrueVtxX);
-      TMSTrueTree->SetBranchAddress("TrueVtxY",                       _TrueVtxY);
-      TMSTrueTree->SetBranchAddress("TrueVtxZ",                       _TrueVtxZ);
-      TMSTrueTree->SetBranchAddress("RecoTrackPrimaryParticleVtxId",  _RecoTrueVtxId);
-      TMSTrueTree->SetBranchAddress("RecoTrackPrimaryParticleIndex",  _RecoTruePartId);
+      // Truth_Info is only used to map reco tracks to truth-particle indices.
+      TMSTrueTree->SetBranchAddress("RecoTrackPrimaryParticleIndex",   _RecoTruePartId);
       TMSTrueTree->SetBranchAddress("RecoTrackSecondaryParticleIndex", _RecoTruePartIdSec);
 
-      TMSTrueSpill->SetBranchAddress("VertexID",                       _TrueVtxId);
-      TMSTrueSpill->SetBranchAddress("TrueVtxN",                       &_TrueVtxN);
-      TMSTrueSpill->SetBranchAddress("RunNo",                          &_TrueRunNo);
+      TMSTrueSpill->SetBranchAddress("SpillNo",                        &_TruthSpillSpillNo);
+      TMSTrueSpill->SetBranchAddress("RunNo",                          &_TruthSpillRunNo);
+      TMSTrueSpill->SetBranchAddress("nTrueParticles",                 &_TruthSpillNTrueParticles);
+      TMSTrueSpill->SetBranchAddress("VertexID",                       _TruthSpillParticleVertexID);
+      TMSTrueSpill->SetBranchAddress("Parent",                         _TruthSpillParent);
+      TMSTrueSpill->SetBranchAddress("BirthPosition",                  _TruthSpillBirthPosition);
+      TMSTrueSpill->SetBranchAddress("TrueVtxN",                       &_TruthSpillTrueVtxN);
+      TMSTrueSpill->SetBranchAddress("TrueVtxID",                      _TruthSpillTrueVtxID);
+      TMSTrueSpill->SetBranchAddress("TrueVtxX",                       _TruthSpillTrueVtxX);
+      TMSTrueSpill->SetBranchAddress("TrueVtxY",                       _TruthSpillTrueVtxY);
+      TMSTrueSpill->SetBranchAddress("TrueVtxZ",                       _TruthSpillTrueVtxZ);
 
     } else {
       fTMSRecoFile = NULL;
@@ -106,16 +110,109 @@ namespace cafmaker
     fTMSRecoFile = NULL;
   }
 
-  unsigned long TMSRecoBranchFiller::ResolveTrueInteractionID(const TruthMatcher * truthMatch,
-                                                              int trueVtxId,
-                                                              double trueVtxX,
-                                                              double trueVtxY,
-                                                              double trueVtxZ) const
+  void TMSRecoBranchFiller::LoadTruthSpillEntry(int spillNo) const
+  {
+    if (fTruthSpillEntryBySpillNo.empty())
+    {
+      for (Long64_t entry = 0; entry < TMSTrueSpill->GetEntries(); ++entry)
+      {
+        TMSTrueSpill->GetEntry(entry);
+        auto [it, inserted] = fTruthSpillEntryBySpillNo.emplace(_TruthSpillSpillNo, entry);
+        if (!inserted && it->second != entry)
+        {
+          std::stringstream ss;
+          ss << "Truth_Spill has multiple entries for SpillNo " << _TruthSpillSpillNo << "\n";
+          throw std::runtime_error(ss.str());
+        }
+      }
+    }
+
+    auto it = fTruthSpillEntryBySpillNo.find(spillNo);
+    if (it == fTruthSpillEntryBySpillNo.end())
+    {
+      std::stringstream ss;
+      ss << "Could not find Truth_Spill entry for SpillNo " << spillNo << "\n";
+      throw std::runtime_error(ss.str());
+    }
+
+    TMSTrueSpill->GetEntry(it->second);
+  }
+
+  unsigned long TMSRecoBranchFiller::ResolveTrueInteractionIDFromVertexIndex(const TruthMatcher * truthMatch, int trueVtxIdx) const
   {
     if (!truthMatch)
       throw std::runtime_error("TMSRecoBranchFiller requires TruthMatcher to resolve true interaction IDs");
+    if (trueVtxIdx < 0 || trueVtxIdx >= _TruthSpillTrueVtxN)
+      throw std::runtime_error("Requested Truth_Spill vertex index is out of range");
 
-    return truthMatch->ResolveVertexID(static_cast<unsigned int>(trueVtxId), trueVtxX, trueVtxY, trueVtxZ);
+    return truthMatch->ResolveVertexID(static_cast<unsigned int>(_TruthSpillTrueVtxID[trueVtxIdx]),
+                                       _TruthSpillTrueVtxX[trueVtxIdx],
+                                       _TruthSpillTrueVtxY[trueVtxIdx],
+                                       _TruthSpillTrueVtxZ[trueVtxIdx]);
+  }
+
+  unsigned long TMSRecoBranchFiller::ResolveRecoTrackInteractionID(const TruthMatcher * truthMatch, int recoTrackIdx) const
+  {
+    if (recoTrackIdx < 0)
+      throw std::runtime_error("Requested reco track index is negative");
+
+    int particleIdx = _RecoTruePartId[recoTrackIdx];
+    if (particleIdx < 0 || particleIdx >= _TruthSpillNTrueParticles)
+    {
+      std::stringstream ss;
+      ss << "Reco track " << recoTrackIdx << " maps to invalid Truth_Spill particle index " << particleIdx << "\n";
+      throw std::runtime_error(ss.str());
+    }
+
+    while (_TruthSpillParent[particleIdx] != -1)
+    {
+      int parentIdx = _TruthSpillParent[particleIdx];
+      if (parentIdx < 0 || parentIdx >= _TruthSpillNTrueParticles)
+      {
+        std::stringstream ss;
+        ss << "Truth_Spill particle " << particleIdx << " has out-of-range parent index " << parentIdx << "\n";
+        throw std::runtime_error(ss.str());
+      }
+      particleIdx = parentIdx;
+    }
+
+    const int trueVtxId = _TruthSpillParticleVertexID[particleIdx];
+    const double birthX = _TruthSpillBirthPosition[particleIdx][0];
+    const double birthY = _TruthSpillBirthPosition[particleIdx][1];
+    const double birthZ = _TruthSpillBirthPosition[particleIdx][2];
+
+    constexpr double kVertexMatchToleranceMm = 100.0;
+    int matchedVtxIdx = -1;
+    for (int i = 0; i < _TruthSpillTrueVtxN; ++i)
+    {
+      if (_TruthSpillTrueVtxID[i] != trueVtxId)
+        continue;
+
+      const double dx = _TruthSpillTrueVtxX[i] - birthX;
+      const double dy = _TruthSpillTrueVtxY[i] - birthY;
+      const double dz = _TruthSpillTrueVtxZ[i] - birthZ;
+      const double dist2 = dx*dx + dy*dy + dz*dz;
+      if (dist2 > kVertexMatchToleranceMm * kVertexMatchToleranceMm)
+        continue;
+
+      if (matchedVtxIdx >= 0)
+      {
+        std::stringstream ss;
+        ss << "Reco track " << recoTrackIdx << " matches multiple Truth_Spill vertices for vertex ID " << trueVtxId << "\n";
+        throw std::runtime_error(ss.str());
+      }
+      matchedVtxIdx = i;
+    }
+
+    if (matchedVtxIdx < 0)
+    {
+      std::stringstream ss;
+      ss << "Reco track " << recoTrackIdx << " could not match primary particle birth position to a Truth_Spill vertex for vertex ID "
+         << trueVtxId << " within " << kVertexMatchToleranceMm << " mm\n";
+      throw std::runtime_error(ss.str());
+    }
+
+    return ResolveTrueInteractionIDFromVertexIndex(truthMatch, matchedVtxIdx);
   }
 
   // here we copy all the TMS reco into the SRTMS branch of the StandardRecord object.
@@ -148,16 +245,11 @@ namespace cafmaker
 
     caf::SRTMSInt *interaction;
 
-    // Fill Truth parts first?
-    for (int i_tru=0; i_tru< TMSTrueSpill->GetEntries(); i_tru++)
+    LoadTruthSpillEntry(LastSpillNo);
+    for (int i_tvtx = 0; i_tvtx < _TruthSpillTrueVtxN; ++i_tvtx)
     {
-      TMSTrueSpill->GetEntry(i_tru);
-
-      for (int i_tvtx=0; i_tvtx<_TrueVtxN; i_tvtx++)
-      {
-        auto neutrino_event_id = ResolveTrueInteractionID(truthMatcher, _TrueVtxId[i_tvtx], _TrueVtxX[i_tvtx], _TrueVtxY[i_tvtx], _TrueVtxZ[i_tvtx]);
-        (void)truthMatcher->GetTrueInteraction(sr, neutrino_event_id, true); // called for side effect of registering the interaction; cast suppresses unused-return-value warning
-      }
+      auto neutrino_event_id = ResolveTrueInteractionIDFromVertexIndex(truthMatcher, i_tvtx);
+      (void)truthMatcher->GetTrueInteraction(sr, neutrino_event_id, true); // called for side effect of registering the interaction; cast suppresses unused-return-value warning
     }
 
     // the i index is incremented at the end of the following while()
@@ -196,7 +288,7 @@ namespace cafmaker
           /*  Fill Truth
            *  The run numbers in the GHEP(?) or edep files are of the run number, followed by the event number, so we recreate that.
            * Long cos it's very long innit. Sorry. */
-          const auto genieIxnID = ResolveTrueInteractionID(truthMatcher, _RecoTrueVtxId[j], _TrueVtxX[j], _TrueVtxY[j], _TrueVtxZ[j]);
+          const auto genieIxnID = ResolveRecoTrackInteractionID(truthMatcher, j);
           caf::SRTrueInteraction& srTrueInt = truthMatcher->GetTrueInteraction(sr, genieIxnID);
 
           const int srTrueIntIdx = static_cast<int>(std::distance(sr.mc.nu.begin(),
@@ -228,10 +320,9 @@ namespace cafmaker
   void TMSRecoBranchFiller::FillInteractions(const TruthMatcher * truthMatch, caf::StandardRecord &sr) const
   {
 
-    for (int i_int = 0; i_int<_TrueVtxN; i_int++)
+    for (int i_int = 0; i_int < _TruthSpillTrueVtxN; ++i_int)
     {
-      //Long_t neutrino_event_id = _TrueVtxId[i_int];//mc_int_edepsimId[i_int];
-      auto neutrino_event_id = ResolveTrueInteractionID(truthMatch, _TrueVtxId[i_int], _TrueVtxX[i_int], _TrueVtxY[i_int], _TrueVtxZ[i_int]);
+      auto neutrino_event_id = ResolveTrueInteractionIDFromVertexIndex(truthMatch, i_int);
       caf::SRTrueInteraction & srTrueInt = truthMatch->GetTrueInteraction(sr, neutrino_event_id);
       LOG.VERBOSE() << "    --> resulting SRTrueInteraction has the following particles in it:\n";
       for (const caf::SRTrueParticle & part : srTrueInt.prim)
--- a/src/reco/TMSRecoBranchFiller.cxx
+++ b/src/reco/TMSRecoBranchFiller.cxx
@@ -455,7 +455,7 @@ namespace cafmaker
 
     for (const Trigger & trigger : fTriggers)
     {
-      if (triggerType < 0 || triggerType == fTriggers.back().triggerType)
+      if (triggerType < 0 || triggerType == trigger.triggerType)
       {
         triggers.push_back(trigger);
       }
--- a/src/reco/TMSRecoBranchFiller.cxx
+++ b/src/reco/TMSRecoBranchFiller.cxx
@@ -82,6 +82,7 @@ namespace cafmaker
       TMSTrueSpill->SetBranchAddress("nTrueParticles",                 &_TruthSpillNTrueParticles);
       TMSTrueSpill->SetBranchAddress("VertexID",                       _TruthSpillParticleVertexID);
       TMSTrueSpill->SetBranchAddress("Parent",                         _TruthSpillParent);
+      TMSTrueSpill->SetBranchAddress("TrackId",                        _TruthSpillTrackID);
       TMSTrueSpill->SetBranchAddress("BirthPosition",                  _TruthSpillBirthPosition);
       TMSTrueSpill->SetBranchAddress("TrueVtxN",                       &_TruthSpillTrueVtxN);
       TMSTrueSpill->SetBranchAddress("TrueVtxID",                      _TruthSpillTrueVtxID);
@@ -151,12 +152,12 @@ namespace cafmaker
                                                          _TruthSpillTrueVtxZ[trueVtxIdx]);
   }
 
-  unsigned long TMSRecoBranchFiller::ResolveRecoTrackInteractionID(const TruthMatcher * truthMatch, int recoTrackIdx) const
+  int TMSRecoBranchFiller::ResolveRecoTrackTruthParticleIndex(int recoTrackIdx) const
   {
     if (recoTrackIdx < 0)
       throw std::runtime_error("Requested reco track index is negative");
 
-    int particleIdx = _RecoTruePartId[recoTrackIdx];
+    const int particleIdx = _RecoTruePartId[recoTrackIdx];
     if (particleIdx < 0 || particleIdx >= _TruthSpillNTrueParticles)
     {
       std::stringstream ss;
@@ -164,17 +165,54 @@ namespace cafmaker
       throw std::runtime_error(ss.str());
     }
 
+    return particleIdx;
+  }
+
+  int TMSRecoBranchFiller::FindTruthSpillParticleIndex(int vertexId, int trackId) const
+  {
+    for (int i = 0; i < _TruthSpillNTrueParticles; ++i)
+    {
+      if (_TruthSpillParticleVertexID[i] == vertexId && _TruthSpillTrackID[i] == trackId)
+        return i;
+    }
+    return -1;
+  }
+
+  int TMSRecoBranchFiller::ResolvePrimaryTruthParticleIndex(int particleIdx, int recoTrackIdx) const
+  {
+    std::vector<int> visited;
     while (_TruthSpillParent[particleIdx] != -1)
     {
-      int parentIdx = _TruthSpillParent[particleIdx];
-      if (parentIdx < 0 || parentIdx >= _TruthSpillNTrueParticles)
+      if (std::find(visited.begin(), visited.end(), particleIdx) != visited.end())
       {
-        std::stringstream ss;
-        ss << "Truth_Spill particle " << particleIdx << " has out-of-range parent index " << parentIdx << "\n";
-        throw std::runtime_error(ss.str());
+        LOG.WARNING() << "TMS reco track " << recoTrackIdx
+                      << " encountered a loop while resolving Truth_Spill parents;"
+                      << " falling back to particle index " << particleIdx << "\n";
+        return particleIdx;
+      }
+      visited.push_back(particleIdx);
+
+      const int parentTrackId = _TruthSpillParent[particleIdx];
+      const int vertexId = _TruthSpillParticleVertexID[particleIdx];
+      const int parentIdx = FindTruthSpillParticleIndex(vertexId, parentTrackId);
+      if (parentIdx < 0)
+      {
+        LOG.WARNING() << "TMS reco track " << recoTrackIdx
+                      << " could not resolve parent track ID " << parentTrackId
+                      << " from Truth_Spill particle index " << particleIdx
+                      << " (vertex ID " << vertexId << ");"
+                      << " falling back to the current particle instead of crashing\n";
+        return particleIdx;
       }
       particleIdx = parentIdx;
     }
+
+    return particleIdx;
+  }
+
+  unsigned long TMSRecoBranchFiller::ResolveRecoTrackInteractionID(const TruthMatcher * truthMatch, int recoTrackIdx) const
+  {
+    const int particleIdx = ResolvePrimaryTruthParticleIndex(ResolveRecoTrackTruthParticleIndex(recoTrackIdx), recoTrackIdx);
 
     const int trueVtxId = _TruthSpillParticleVertexID[particleIdx];
     const double birthX = _TruthSpillBirthPosition[particleIdx][0];
@@ -297,7 +335,8 @@ namespace cafmaker
                                                                                { return ixn.id == srTrueInt.id; })));
 
           // TODO: Make TMS care about prim/sec tracks (check _RecoTruePartIdSec for secondaries)
-          const int partG4ID = _RecoTruePartId[j];
+          const int recoTruthParticleIdx = ResolveRecoTrackTruthParticleIndex(j);
+          const int partG4ID = _TruthSpillTrackID[recoTruthParticleIdx];
           truthMatcher->GetTrueParticle(sr, srTrueInt, partG4ID, true);
           const int truthVecIdx = static_cast<int>(std::distance(srTrueInt.prim.begin(),
                                                                  std::find_if(srTrueInt.prim.begin(), srTrueInt.prim.end(),
--- a/src/reco/TMSRecoBranchFiller.cxx
+++ b/src/reco/TMSRecoBranchFiller.cxx
@@ -412,35 +412,43 @@ namespace cafmaker
 
         lastSpillNo = _SpillNo;
 
-        Trigger & prev_trig = fTriggers.back(); // trigger before 'trig'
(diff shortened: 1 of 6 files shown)


jdkio:
There are test files in `/exp/dune/data/users/kleykamp/tests/fixed_select/`. @mt82, can you run the plots and see if you still see the error?

mt82:
Hi @jdkio, this is what I did:

<details>
  <summary>installed ND_CAFMaker using this script</summary>

``` bash
#!/usr/bin/env bash

if [ -n "${APPTAINER_CONTAINER:-}" ]; then
    echo "[INFO] Inside container"

    ROOTFOLDER=$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)
    
    source /cvmfs/dune.opensciencegrid.org/products/dune/setup_dune.sh

    ##################################
    # Now build CAFMaker itself
    ##################################
        
    DIRECTORY=ND_CAFMaker
    BRANCH=jdkio/fix_truth_matcher

    cd ${ROOTFOLDER}
    touch ${DIRECTORY}
    rm -rf ${DIRECTORY}
    
    git clone -b ${BRANCH} https://github.com/DUNE/ND_CAFMaker.git
    cd ND_CAFMaker/
    ./build_deps.sh
    source ndcaf_setup.sh debug
    make
else
    /cvmfs/oasis.opensciencegrid.org/mis/apptainer/current/bin/apptainer exec -B /cvmfs,/exp,/nashome,/pnfs/dune,/opt,/run/user,/etc/hostname,/etc/hosts,/etc/krb5.conf --ipc --pid /cvmfs/singularity.opensciencegrid.org/fermilab/fnal-dev-sl7:latest bash "$0"
fi
```
</details>

<details>
  <summary>ran ND_CAFMaker with the following script and input</summary>

##### script
``` bash
#!/usr/bin/env bash

if [ -n "${APPTAINER_CONTAINER:-}" ]; then
    echo "[INFO] Inside container"

    ROOTFOLDER=$(cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd)

    cd ${ROOTFOLDER}/ND_CAFMaker/
    source ndcaf_setup.sh debug
    cd ..
    mkdir -p ana
    ND_CAFMaker/bin/makeCAF jacobCAF.fcl
else
    /cvmfs/oasis.opensciencegrid.org/mis/apptainer/current/bin/apptainer exec -B /cvmfs,/exp,/nashome,/pnfs/dune,/opt,/run/user,/etc/hostname,/etc/hosts,/etc/krb5.conf --ipc --pid /cvmfs/singularity.opensciencegrid.org/fermilab/fnal-dev-sl7:latest bash "$0"
fi
```

##### input
```bash
#include "NDCAFMaker.fcl"

nd_cafmaker: @local::standard_nd_cafmaker

nd_cafmaker.CAFMakerSettings.GHEPFiles: [
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024640.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024641.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024642.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024643.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024644.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024645.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024646.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024647.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024648.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024649.GHEP.root",

   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024640.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024641.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024642.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024643.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024644.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024645.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024646.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024647.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024648.GHEP.root",
   "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024649.GHEP.root"
]

nd_cafmaker.CAFMakerSettings.OutputFile: "ana/jacobCAF.root"
nd_cafmaker.CAFMakerSettings.NDLArRecoFile: "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-mlreco/MicroProdN4p1_NDComplex_FHC.spine.full.light/MLRECO_SPINE/0002000/MicroProdN4p1_NDComplex_FHC.spine.full.light.0002464.MLRECO_SPINE.hdf5"
nd_cafmaker.CAFMakerSettings.TMSRecoFile: "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-tms-reco/MicroProdN4p1_NDComplex_FHC.tmsreco.full/TMSRECO/0002000/0002400/MicroProdN4p1_NDComplex_FHC.tmsreco.full.0002464.TMSRECO.root"
nd_cafmaker.CAFMakerSettings.EdepsimFile: "/pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-spill-build/MicroProdN4p1_NDComplex_FHC.spill.full/EDEPSIM_SPILLS/0002000/0002400/MicroProdN4p1_NDComplex_FHC.spill.full.0002464.EDEPSIM_SPILLS.root"
```
</details>

This is the output I obtained:

```
./run-ND_CAFMaker.sh 
[INFO] Inside container
Setting up larsoft UPS area... /cvmfs/larsoft.opensciencegrid.org
Setting up DUNE UPS area... /cvmfs/dune.opensciencegrid.org/products/dune/
1780569847 NOTICE Messenger : [s] <SetPrioritiesFromXmlFile (136)> : Reading msg stream priorities from XML file: /cvmfs/larsoft.opensciencegrid.org/products/genie/v3_04_00i/Linux64bit+3.10-2.17-e26-debug/GENIE-Generator/config/Messenger.xml
1780569847 NOTICE Messenger : [s] <SetPrioritiesFromXmlFile (136)> : Reading msg stream priorities from XML file: /cvmfs/larsoft.opensciencegrid.org/products/genie/v3_04_00i/Linux64bit+3.10-2.17-e26-debug/GENIE-Generator/config/Messenger_whisper.xml
Error in <TTree::Branch>: The class requested (vector<vector<vector<unsigned long> > >) for the branch "geoEffThrowResults" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<vector<vector<unsigned long> > >) to avoid to write corrupted data.
Filling Reco info for the following cases:
   ND-LAr (Deep-Learn-Physics ML)
   TMS
   ND-LAr + TMS matching
GTreeContainer INFO: Loaded TTree for run 24640 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024640.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24641 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024641.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24642 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024642.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24643 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024643.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24644 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024644.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24645 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024645.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24646 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024646.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24647 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024647.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24648 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024648.GHEP.root
GTreeContainer INFO: Loaded TTree for run 24649 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.ndlarfid.0024649.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024640 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024640.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024641 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024641.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024642 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024642.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024643 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024643.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024644 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024644.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024645 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024645.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024646 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024646.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024647 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024647.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024648 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024648.GHEP.root
GTreeContainer INFO: Loaded TTree for run 1000024649 from file: /pnfs/dune/persistent/physicsgroups/dunendsim/abooth/nd-production/MicroProdN4p1/run-genie/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid/GHEP/0024000/0024600/MicroProdN4p1_NDComplex_FHC.genie.rockantindlarfid.0024649.GHEP.root
Processing 13 triggers...
[                                                          ]GTreeContainer FATAL: GENIE event number 3682 was not found in run 24645
terminate called after throwing an instance of 'std::range_error'
  what():  GENIE event number 3682 was not found in run 24645

./run-ND_CAFMaker.sh: line 15:  1730 Aborted                 (core dumped) ND_CAFMaker/bin/makeCAF jacobCAF.fcl
```

I’m still getting an error and I’m not sure if I’m missing something

mt82:
Regarding the file you created at `/exp/dune/data/users/kleykamp/tests/fixed_select/example-CAF.root`, it looks OK (i.e. no peaks).

nu_E_vs_vtx_Z

jdkio:
@mt82, this is an issue with the TMS files. They only save run number once rather than per vertex, so using those run numbers isn't going to disambiguate between a rock file and a fiducial file. I'm looking at what can be done about it but it's not going to be clean fix

My test sample didn't have a TMS file so it didn't crash. It's good to see that the fix seems to have worked

mt82:
Thanks @jdkio for taking ownership of this issue.
I just wanted to check whether I’ve understood correctly.
The changes you made to the code above are meant to handle the case where the eventId variable does not match the entry number. Am I understanding this correctly?
However, I don’t understand why this solution doesn’t work for TMS files that were already generated (up4.1).
Is there any other convention or assumption that the code relies on?
Is this because the TMS code currently assumes that the run number is the same across the entire spill, an assumption that no longer holds with the addition of rock muons?
In any case, thanks again for the work you’ve done. The changes look good to me.

jdkio:
@mt82, I found the issue in the TMS truth-matching path. The TMS files save one run number per spill, but the truth matching assumes access to stable `(run number, vertex id)` information that can be used to recover the originating GENIE event. In these cases, that assumption breaks down, so we can’t recover the correct source GENIE information from the TMS truth info alone.

### Fix
The updated branch resolves the interaction in the TMS-specific path using the **truth vertex position** from `Truth_Spill` instead of relying on `EventId` alone. Concretely, it builds a lookup from true vertex position in the edep-sim file back to the packed CAF/EDepSim vertex ID, and uses that to recover the intended truth association.

I see this as a temporary fix. Longer-term, we should agree on a stable event-identification convention for DUNE and then update the TMS path to use that directly.

### Performance
I also added cached and binned lookup for the position-based resolver. For `ResolveVertexIDFromRunAndPosition()` on the tested sample:

- without caching: `7098` calls, `~941 ms` total
- with caching: `7098` calls, `~65–73 ms` total
- the y-binning appears to reduce candidate comparisons by about a factor of 6

You can print the timing summary at the end with:
```bash
export ND_CAFMAKER_TIMING=1
```

mt82:
Hi @jdkio. Thank you again for working on this. I haven’t gone through the code in detail yet, because I first tried running the test I mentioned in the previous [comment](https://github.com/DUNE/ND_CAFMaker/pull/151#issuecomment-4621698643).

The code now passes the point where it used to stop before your changes, which confirms that your modifications correctly handle the problematic situation that originally triggered this issue.

However, the code still seems to throw an exception at a later stage. The exception is:

```
what(): Truth_Spill particle 2729 has out-of-range parent index 14000
```

and it is thrown at line `174` of `TMSRecoBranchFiller.cxx`.

The attached figure shows additional details.

I’m absolutely available to support you in any way you think might be helpful.

image

jdkio:
@mt82, you're running run 2464, right? I tested with run 2460. I'll try your run and see if I get the same crash. I've attached my test fcl. Can you please check that yours is the same other than the run numbers?

[ndcafmakerjobtesttms.fcl.txt](https://github.com/user-attachments/files/28845607/ndcafmakerjobtesttms.fcl.txt)

Edit: I also get the crash. Investigating

jdkio:
@mt82, I think it's fixed now. I think there are some edge cases where the TMS code doesn't save the information correctly. So I did my best to fail gracefully here instead of just crashing. I think the TMS code needs a longer-term fix, which I can take a look at doing now

mt82:
Fantastic, @jdkio! Just to confirm, with your version ND_CAFMaker successfully reaches the end of the process

alexbooth92:
Excellent news @jdkio, thank you for your work on this!

alexbooth92 (approved):
In the interest of getting this merged, I've made an issue to track discussions / progress towards revisiting event mapping (#157). I've also added a reference to removing `ResolveTrueInteractionIDFromVertexIndex` in a subissue #158 so that we don't forget about this.

src/truth/FillTruth.cxx
```diff
@@ -31,6 +33,47 @@
...
+  {
+    static const bool enabled = []()
+    {
+      const char * env = std::getenv("ND_CAFMAKER_TIMING");
+      if (!env) return false;
+      const std::string value(env);
+      return !value.empty() && value != "0" && value != "false" && value != "FALSE";
+    }();
+    return enabled;
+  }
```
chenel: Surely an environment variable is not the best path here.  Is there some reason this couldn't be a FCL parameter?

src/truth/FillTruth.cxx
```diff
@@ -31,6 +33,47 @@
...
+
+  class ScopedTiming
```
chenel: Two things here.

* This class is sort of mysterious at first glance.  I guess the idea is that it counts time spent only inside a certain function or context?  Some doxygen commentary would help a lot.
* This seems like (after some cleanup) it would probably go better as a standalone in the `src/util` folder.  I don't think it makes a lot of sense to bury it in the `.cxx` here.  See overall comment.

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -1,6 +1,59 @@
...
+  {
+    static const bool enabled = []()
+    {
+      const char * env = std::getenv("ND_CAFMAKER_TIMING");
+      if (!env) return false;
+      const std::string value(env);
+      return !value.empty() && value != "0" && value != "false" && value != "FALSE";
+    }();
+    return enabled;
+  }
```
chenel: Similar comment here: why are we using environment vars and not FCL params?

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -106,12 +217,196 @@ namespace cafmaker
...
+
+  int TMSRecoBranchFiller::ResolvePrimaryTruthParticleIndex(int particleIdx, int recoTrackIdx) const
+  {
+    ScopedTiming timing(fTiming.enabled, fTiming.resolvePrimaryTruthParticleIndex);
+
+    std::vector<int> visited;
+    while (_TruthSpillParent[particleIdx] != -1)
+    {
+      ++fTiming.resolvePrimaryParentSteps;
+      if (std::find(visited.begin(), visited.end(), particleIdx) != visited.end())
```
chenel: if you used a `std::unordered_set<int>` for `visisted` above, then this could be a `visited.find()` call, which reduces an O(N^2) worst-case lookup to an O(N) one.  not a big deal, but an easy fix

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -106,12 +217,196 @@ namespace cafmaker
...
+      }
+    }
+    return -1;
+  }
+
+  int TMSRecoBranchFiller::ResolvePrimaryTruthParticleIndex(int particleIdx, int recoTrackIdx) const
+  {
+    ScopedTiming timing(fTiming.enabled, fTiming.resolvePrimaryTruthParticleIndex);
+
+    std::vector<int> visited;
```
chenel: might as well move this declaration inside the `while`, to avoid an unnecessary allocation.  also, consider making it a `std::unordered_set<>` per comment below

src/truth/FillTruth.h
```diff
@@ -179,6 +183,12 @@
       /// \param createNew  Should a new SRTrueInteraction be made if one corresponding to the given ID is not found?
       /// \return           The caf::SRTrueParticle that was found, or if none found and createNew is true, a new instance
       caf::SRTrueInteraction & GetTrueInteraction(caf::StandardRecord & sr, unsigned long ixnID, bool createNew = true) const;
+      /// Resolve an EDepSim EventId plus vertex position (mm) into the legacy packed vertex ID
+      /// (run*1e6 + event). This preserves current CAFMaker conventions, even though the encoding is brittle.
+      unsigned long ResolveVertexID(unsigned int evtNum, double x, double y, double z) const;
+      /// TMS-specific resolver: use the Truth_Spill vertex position to recover the legacy packed
+      /// vertex ID. The base run number is retained only for diagnostics; matching is position-only.
+      unsigned long ResolveVertexIDFromRunAndPosition(unsigned long baseRunNum, double x, double y, double z) const;
```
chenel: I don't like `baseRunNum` here as a name overmuch.  `runNumForErrMsg` would be a lot cleaner (also request adjusting the docstring)

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -245,42 +563,53 @@ namespace cafmaker
...
+                        << ", triggerTime_s=" << trig.triggerTime_s
+                        << ", triggerTime_ns=" << trig.triggerTime_ns
+                        << "\n";
+        }
+        else
+        {
+          fTriggers.emplace_back();
+          Trigger & trig = fTriggers.back();
+          trig.evtID = entry;
+          trig.triggerType = 1; // TODO real number?
```
chenel: This code is duplicated from the primary branch of the `if` -- probably worth just extracting it above the `if`?

src/reco/TMSRecoBranchFiller.h
```diff
@@ -34,7 +37,54 @@ namespace cafmaker
 
       ~TMSRecoBranchFiller();
 
+      struct TimingSummary
+      {
+        double totalMs = 0.0;
+        std::size_t calls = 0;
+      };
+
+      struct TimingStats
```
chenel: This class doesn't belong in the header, it should be in the .cxx.  It's implementation details only.

src/truth/FillTruth.cxx
```diff
@@ -612,21 +690,276 @@ namespace cafmaker
...
+    auto it = fEventToVertexIDs.find(evtNum);
+    if (it == fEventToVertexIDs.end())
+    {
+      std::stringstream ss;
+      ss << "EDepSim event ID " << evtNum << " was not found while resolving a packed vertex ID\n";
+      LOG.FATAL() << ss.str();
+      throw std::range_error(ss.str());
+    }
+
+    constexpr double kPositionToleranceMm = 1.0;
```
chenel: this is a physicsy parameter whose value affects the matching performance.  I would strongly recommend this get wired up as a FCL parameter instead of a compile-time constant.

src/truth/FillTruth.cxx
```diff
@@ -612,21 +690,276 @@ namespace cafmaker
...
+    const auto cacheKey = std::make_tuple(x, y, z);
+    auto cacheIt = fResolvedVertexIDByPosition.find(cacheKey);
+    if (cacheIt != fResolvedVertexIDByPosition.end())
+    {
+      if (fTiming.enabled) ++fTiming.resolveRunPosCacheHits;
+      return cacheIt->second;
+    }
+    if (fTiming.enabled) ++fTiming.resolveRunPosCacheMisses;
+
+    constexpr double kPositionToleranceMm = 1.0;
```
chenel: same comment as on the previous instance of `kPositionToleranceMm` in the previous function---would be much better as a FCL parameter.  Probably you could use the same one for both.

src/reco/TMSRecoBranchFiller.cxx
```diff
@@ -106,12 +217,196 @@ namespace cafmaker
...
+    ScopedTiming timing(fTiming.enabled, fTiming.resolveRecoTrackInteractionID);
+
+    const int particleIdx = ResolvePrimaryTruthParticleIndex(ResolveRecoTrackTruthParticleIndex(recoTrackIdx), recoTrackIdx);
+
+    const int trueVtxId = _TruthSpillParticleVertexID[particleIdx];
+    const double birthX = _TruthSpillBirthPosition[particleIdx][0];
+    const double birthY = _TruthSpillBirthPosition[particleIdx][1];
+    const double birthZ = _TruthSpillBirthPosition[particleIdx][2];
+
+    constexpr double kVertexMatchToleranceMm = 100.0;
```
chenel: this is a physicsy parameter whose value affects the performance.  I would strongly recommend this get wired up as a FCL parameter instead of a compile-time constant.

src/truth/FillTruth.h
```diff
@@ -9,12 +9,16 @@
 #define ND_CAFMAKER_FILLTRUTH_H
 
 #include <cmath>
+#include <cstddef>
 #include <functional>
 #include <iostream>
 #include <limits>
+#include <cmath>
```
chenel: whoops, this is a duplicate, it's already included above

src/reco/TMSRecoBranchFiller.h
```diff
@@ -48,52 +98,65 @@ namespace cafmaker
...
+
+      float _TMSStartTime[kTMSMaxTracks];
+      float _TrackTime[kTMSMaxTracks];
+
+      float _DirectionX_Downstream[kTMSMaxTracks];
+      float _DirectionZ_Downstream[kTMSMaxTracks];
+      float _DirectionX_Upstream[kTMSMaxTracks];
+      float _DirectionZ_Upstream[kTMSMaxTracks];
+
+      float _TrackHitPos[kTMSMaxTracks][kTMSMaxLineHits][3];     ///< Kalman filtered reco. track hit positions (x,y,z)
```
chenel: The innermost array here used to have size 4 (including t).  Is this going to make us unable to read some older generation of TMS files?  If so, what kind of warning do we need to make sure we don't get bitten by it in future?

chenel (approved):
I'm glad to see you've tracked down the issue.  With all the timing infrastructure, this is quite a lengthy PR that has a lot of machinery, which means I'm not totally sure I understand everything that's happening.  But as far as I understood it, the overall approach seems right, and I'm glad the testing has borne that out.

I do have some concerns about the implementation, however.  

### ScopedTiming
The vast majority of my concerns are about how the performance tracking stuff was implemented and where it lives.  Overall, I think  the major problem is that you've implemented essentially the same `ScopedTiming` class in two places with different interfaces.  Neither of these places is the right place---this kind of generic tool belongs in `util` as a standalone class, not inside anonymous namespaces in the implementation files for other classes.  `TimingSummary` should go along for the ride and land in `util` as well.  Then I think the generic version probably can keep the TMS version's interface: `ScopedTiming(bool, TimingSummary&)`---seems like the `double&/size_t&` variant was a workaround to avoid coupling `TimingSummary` into `FillTruth`; but that's not a problem once everything is consolidated.

At that point, then I think the `TimingStats` class can (and should) live in the `.cxx` and be used in a private context only.  It seems like it's only in the public header file now so that `ScopedTiming` can talk to it.  But we addressed that above. 

### TMSRecoBranchFiller complexity
The other thing you might want to consider is breaking a lot of this new functionality into a helper class---`TMSTruthResolver`?---because `TMSRecoBranchFiller` is getting pretty complex.  This would not be trivial because of the need to be careful with the ROOT  `SetBranchAddress()` stuff.  But maybe it's not that hard.  I won't _insist_ on it for this PR to merge, but it's worth considering if you have the bandwidth.    If not, at least adding to @alexbooth92's #157 is worthwhile.

chenel:
Note that I clicked "approve" rather than "request changes" so that the conveners (@alexbooth92 @mt82) can help judge whether my comments need addressing now, or whether we need to press on.  I really would prefer the `ScopedTimer`/`TimingSummary` mess to get resolved ASAP, but I will not die on that hill.

src/reco/TMSRecoBranchFiller.h
jdkio: This array is filled with kalman hits despite the name. That branch is a three vector, not four:
`Reco_Tree->Branch("KalmanPos",      RecoTrackKalmanPos,       "KalmanPos[nTracks][200][3]/F");`
That said, we might want to change that in TMS, so it might be safer to keep this a 4 vector. Not sure

src/truth/FillTruth.h
jdkio: thanks

alexbooth92:
Thanks for the detailed comments, Jeremy. My view is that this can be merged so that we can press on with a CAF respin, provided that two new issues, one related to env var -> fhicl param and one related to `ScopedTimer/TimingSummary`, are created.

src/reco/TMSRecoBranchFiller.h
jdkio: Made this a 4 vector for future compatibility's sake

src/reco/TMSRecoBranchFiller.cxx
jdkio: Made this a fcl param

src/truth/FillTruth.cxx
jdkio: Made this a fcl param

src/truth/FillTruth.cxx
jdkio: Made this a fcl param

src/reco/TMSRecoBranchFiller.cxx
jdkio: Good idea

src/truth/FillTruth.cxx
jdkio: Removing timing stuff, see comment

src/reco/TMSRecoBranchFiller.cxx
jdkio: Removing timing stuff, see comment

src/reco/TMSRecoBranchFiller.h
jdkio: Removing timing stuff, see comment

src/reco/TMSRecoBranchFiller.cxx
jdkio: Changed it to an unordered set. But we want it outside the while loop

src/truth/FillTruth.cxx
jdkio: Removing timing stuff, see comment

src/truth/FillTruth.h
jdkio: Done

jdkio:
@chenel, thanks for the suggestions. I implemented most of them

I removed the timing information. That stuff is still saved in a branch `jdkio/fix_truth_matcher_with_timing_backup` should we want to explore that again

src/reco/TMSRecoBranchFiller.cxx
jdkio: fixed

jdkio:
@chenel, @alexbooth92, @mt82, let me know if I'm okay to merge

mt82:
Hi @jdkio , thanks again for all the time you dedicated to this issue—I really appreciate it. The test completed successfully, as makeCAF ran to completion without crashing. However, I’ve noticed that the initial processing phase (i.e., reading data from the files, I assume) now seems significantly slower. That said, I don’t have any objections to proceeding with the merge

jdkio:
@mt82, yeah it creates a map between vertexes position and vertex id using the edep sim file so the TMS has enough information to match tracks. I didn't bother optimizing because we should spend that effort actually fixing the TMS. And making the setup more unified in general

PR #160, opened 2026-06-17
130 create a branch filler for sandreco experimental

slanzi00:
Adds a branch filler for SANDReco via template explicit specialization, with separate support paths for sandreco-legacy and sandreco-experimental controlled by compilation flags.

### DIFF
--- a/src/Params.h
+++ b/src/Params.h
@@ -20,7 +20,7 @@ namespace cafmaker
 
     // fixme: placeholder.  won't work for data, which will need to either pass through from upstream or interface to POT database here
     fhicl::Atom<float> POTPerSpill { fhicl::Name("POTPerSpill"), fhicl::Comment("Fixed POT per spill (units of 10^13).")};
-    
+
     fhicl::Atom<bool> fhc        { fhicl::Name("IsFHC"),    fhicl::Comment("Is this an FHC run?"), true };
     fhicl::Atom<bool> IsGasTPC   { fhicl::Name("IsGasTPC"), fhicl::Comment("Was GArTPC geometry used? (If not, TMS)"), false };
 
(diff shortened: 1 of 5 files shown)


slanzi00:
tested on `dunegpvm` with MicroProdN4p1 files

src/Params.h
```diff
@@ -43,9 +43,10 @@ namespace cafmaker
     fhicl::OptionalAtom<std::string> ndlarRecoFile  { fhicl::Name{"NDLArRecoFile"}, fhicl::Comment("Input ND-LAr (ML) reco .h5 file") };
     fhicl::OptionalAtom<std::string> tmsRecoFile  { fhicl::Name{"TMSRecoFile"}, fhicl::Comment("Input TMS reco .root file") };
     fhicl::OptionalAtom<std::string> sandRecoFile  { fhicl::Name{"SANDRecoFile"}, fhicl::Comment("Input SAND reco .root file") };
+    fhicl::OptionalAtom<std::string> sandRecoExperimentalFile { fhicl::Name{"SANDRecoExperimentalFile"}, fhicl::Comment("Input SAND fake_reco CAF .root file") };
```
mt82, line 46: - Here and elsewhere, I would suggest hereafter dropping the "experimental" suffix.
- Are ND_CAFMaker and sandreco outputs aware of which reconstruction algorithms have been used? If not, as I expect, we can safely remove the term "fake_reco" from the field comment

src/Params.h
```diff
@@ -20,7 +20,7 @@ namespace cafmaker
 
     // fixme: placeholder.  won't work for data, which will need to either pass through from upstream or interface to POT database here
     fhicl::Atom<float> POTPerSpill { fhicl::Name("POTPerSpill"), fhicl::Comment("Fixed POT per spill (units of 10^13).")};
-    
+
```
mt82, line 23: Unwanted change?

src/makeCAF.C
```diff
@@ -127,21 +127,33 @@ std::vector<std::unique_ptr<cafmaker::IRecoBranchFiller>> getRecoFillers(const c
...
+    std::cout << "   SAND (legacy)\n";
+  }
+
+  #else
+
+  std::string sandExpFile;
+  if (par().cafmaker().sandRecoExperimentalFile(sandExpFile)) {
+    recoFillers.emplace_back(
+      std::make_unique<cafmaker::SANDRecoBranchFiller<cafmaker::SANDRecoVersion::experimental>>(sandExpFile));
+    std::cout << "   SAND (experimental)\n";
```
mt82, line 152: As above

src/reco/SANDRecoBranchFiller.h
```diff
@@ -1,64 +1,84 @@
...
-  };
+namespace caf {
+class StandardRecord;
 }
 
-#endif
\ No newline at end of file
+namespace cafmaker {
+
+enum class SANDRecoVersion : std::int8_t { legacy, experimental };
```
mt82, line 25: I am not sure version (`SANDRecoVersion`) is the right word. What about variant, flavour, edition?

mt82 (changes requested):
Hi @slanzi00. Thank you very much for the work you did. Overall these changes look good to me — nice work. I’d love to hear @chenel's point of view as well.

Minor comments:
- Mixing formatting-only edits with substantial functional changes makes the diff harder to review; if possible split formatting changes into a separate commit/PR so reviewers can focus on the real logic changes.
- Consider renaming the current "experimental" label to something like "baseline" if it is meant to be the new default/normal path — "experimental" suggests instability.
- I’d avoid using the word "version" to distinguish legacy vs the new path; terms such as "flavour", "edition", or "variant" better convey parallel alternatives rather than implying a strict versioning hierarchy.

Those are all minor; otherwise this looks good to me.

chenel:
I saw this --- I'll have a look at this one after we get the CMake & CI stuff integrated, since it depends on that 😄

PR #163, opened 2026-07-01
Group TMS tracks by time slice

jdkio:
1. **TMS interaction shape fix**
   - `TMSRecoBranchFiller` now makes one `sr.nd.tms.ixn` per TMS `Reco_Tree` row/time slice.
   - Each interaction contains all tracks in that slice as `tracks[j]`, with `ntracks = _nTracks`.

2. **Duplicate match fix**
   - `NDLArTMSUniqueMatchRecoFiller` now calls `Create_matches()` once after collecting all candidate matches, instead of repeatedly inside the TMS-interaction loop.
   - That prevents previously collected candidates from being reconsidered and appended again.

There’s also a small defensive guard around optional TMS truth-overlap use in time matching, but that’s more of a crash-prevention cleanup than a main behavior change.

# Plots of new behavior
Screenshot 2026-06-30 204945

Screenshot 2026-06-30 204958

### DIFF
--- a/src/reco/NDLArTMSUniqueMatchRecoFiller.cxx
+++ b/src/reco/NDLArTMSUniqueMatchRecoFiller.cxx
@@ -226,45 +226,47 @@ namespace cafmaker
       if (use_time) {
         // this handles time-based matching - using truth-level particle times for now instead of light in LAr
         
-	bool timeFail = false;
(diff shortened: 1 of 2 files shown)


chenel (approved):
Looks good from my end.  Thanks!  Would also appreciate a review by a TMS expert, though.

PR #170, opened 2026-07-10
Add entry for version 5.1.1 in CHANGELOG

alexbooth92:


chenel (approved):
This is fine, but we also need to change the tag version in the `CMakeLists.txt`.  I added a commit doing that.

Probably we need a "make a release" checklist:

- [ ] Update the CHANGELOG
- [ ] Change the version string in the top-level `CMakeLists.txt`
- [ ] use the GitHub "draft a release" page to tag a new version

alexbooth92:
> This is fine, but we also need to change the tag version in the `CMakeLists.txt`. I added a commit doing that.
> 
> Probably we need a "make a release" checklist:
> 
> * [x]  Update the CHANGELOG
> * [x]  Change the version string in the top-level `CMakeLists.txt`
> * [x]  use the GitHub "draft a release" page to tag a new version

Great catch, thanks Jeremy! I like the idea of having release checklist. What do you think is the best way to make sure whoever is tagging sees it? Maybe at the top of the CHANGELOG since this is where you should probably always start?

chenel:
> > This is fine, but we also need to change the tag version in the `CMakeLists.txt`. I added a commit doing that.
> > Probably we need a "make a release" checklist:
> > 
> > * [x]   Update the CHANGELOG
> > * [x]   Change the version string in the top-level `CMakeLists.txt`
> > * [x]   use the GitHub "draft a release" page to tag a new version
> 
> Great catch, thanks Jeremy! I like the idea of having release checklist. What do you think is the best way to make sure whoever is tagging sees it? Maybe at the top of the CHANGELOG since this is where you should probably always start?

Yeah, that's probably as good as anywhere.  Unfortunately I don't think GitHub has any mechanism for enforcing it (where "enforcement" just means "forces you to think it through"), so it'll have to be a discipline only, but top-of-`CHANGELOG` is probably a good signpost.

alexbooth92:
> > > This is fine, but we also need to change the tag version in the `CMakeLists.txt`. I added a commit doing that.
> > > Probably we need a "make a release" checklist:
> > > 
> > > * [x]    Update the CHANGELOG
> > > * [x]    Change the version string in the top-level `CMakeLists.txt`
> > > * [x]    use the GitHub "draft a release" page to tag a new version
> > 
> > 
> > Great catch, thanks Jeremy! I like the idea of having release checklist. What do you think is the best way to make sure whoever is tagging sees it? Maybe at the top of the CHANGELOG since this is where you should probably always start?
> 
> Yeah, that's probably as good as anywhere. Unfortunately I don't think GitHub has any mechanism for enforcing it (where "enforcement" just means "forces you to think it through"), so it'll have to be a discipline only, but top-of-`CHANGELOG` is probably a good signpost.

👍 

https://github.com/DUNE/ND_CAFMaker/pull/171

PR #172, opened 2026-07-16
Redesign the H5 read model from "mutable underlying buffer" to "snapshot"

### LINKED ISSUE #69: High Memory consumption of CAFMaker
On the whole production chain, CAFMaker is what seem to be the most memory consuming step, while it mostly consists in reading and storing things from relatively small files to small file. 

There are a lot of calls to GetEntry on a lot of files in the TruthFilling and we just need the Genie Record out of the Genie trees, so one way to optimize the memory consumption could be to set the status of the non-relevant branches to 0 

image

Those are Valgrind Massif output with the current situation:
image

And disabling irrelevant branches of the genie trees:
image

chenel:
The previous HDF5 view layer used to imagine that having immutable views of the `DatasetBuffer` (+ a 'dirty' flag) was enough to keep a clear view.  This worked accidentally, but the buffer could silently be overwritten without the view knowing, which is worse than bad.

There was also a major memory leak (spurred on by the bad design) whereby the memory handed out by the HDF5 calls---whose ownership passes to the caller---was never freed.

This PR does two things:
(1) fixes both the design and the memory leak in one go by making the DatasetBuffer one-to-one with the read from the file, and managing memory using `std::shared_ptr<>`.   now each read from file is "snapshotted" immediately after read and is immutable; when the last instance of the `shared_ptr` is destroyed, the H5 buffer is also cleaned up.  this has the bonus of straightforwardly allowing the results of multiple reads to be floating around if needed (for example, if for some reason the CAFMaker ever needs to know what's in an adjacent event, now it can---at least when pulling from the SPINE file).

(2) adds in the caching mechanism for data product reads that has been a very longstanding `//todo`.  The immutable snapshot model adopted in (1) makes this straightforward: just cache the snapshot (keyed by product type), and just ensure that a product request checks the event number before accepting (and replaces an old event in the cache if one is found).

**This change has not been verified to produce identical CAFs yet: production-grade testing is a MUST before a merge.**


chenel:
To verify no harm was done and that this PR actually did fix the memory leak, I did an A/B test where I took the same set of input files (GENIE, edep-sim, SPINE) and ran the CAFMaker with them in (A) `main` (commit 522c1712) and (B) the tip of this PR's branch.

First, I verified that the contents were identical.  

<details>
<summary>Claude wrote an updated version of the `diff_cafs` script that was able to check them, and the two output CAFs were identical, branch-for-branch and value-for-value.</summary>

```
$ $app/src/dunesoft/ND_CAFMaker/scripts/diff_cafs_uproot --check-schema    
$data/scratch/test.ndlar-h5perf.caf.root                                   
$data/scratch/test.ndlar-h5perf.h5snapshot.caf.root                        
  File A: /exp/dune/data/users/jwolcott/scratch/test.ndlar-h5perf.caf.root   
  (13 records)                                                               
  File B: /exp/dune/data/users/jwolcott/scratch/test.ndlar-h5perf.h5snapshot 
  .caf.root  (13 records)                                                    
                                                                             
  === schema (TStreamerInfo class versions) ===                              
    identical: 97 classes, same versions in both files. ✓                    
    (so a duneanaobj-schema difference is ruled out as the cause)            
                                                                             
  Branch-name sets are identical (511 leaf branches each).                   
                                                                             
  === alignment: POSITIONAL (record i vs record i), 13 records ===           
      (pass --id-branches to align on an event key instead)                  
  Comparing 13 aligned records across 511 common leaf branches.              
                                                                             
  === results ===                                                            
  511 common leaf branches:                                                  
     465 agreed (value comparison)                                           
      46 agreed (raw serialized bytes; uproot couldn't decode these)         
       0 differ                                                              
       0 could not be compared at all                                        
                                                                             
  No differences in any branch that could be compared. ✓

```
</details>

Then, I wanted to study the memory usage.  Because the input files only contain a few files (mine, `MiniProdN5p3_NDComplex_FHC.spine.full.sanddrift.0000003.MLRECO_SPINE.hdf`, had 13), it was hard to see anything against the noise of the overall memory usage from loading events from GENIE and edep-sim.  Claude wrote me [a script](https://github.com/DUNE/ND_CAFMaker/pull/172/changes#diff-a49af721937125ca4d60ab9f68af6ba86ffa3486f165ecda2b51927223b9bf32) that takes an input SPINE file and repeats the triggers in it cyclically as many times as requested.  I did an A/B test with one of these that had 650 triggers (the original 13, repeated 50 times).  Running the CAFMaker on a GPVM with this file took about 9 hours for both A & B.  There, the memory leak (studied with [another script Claude wrote](https://github.com/DUNE/ND_CAFMaker/pull/172/changes#diff-255abffdda763dee03b37b7352b2ef4444b9ac57c5b0668e7a7a13ba7d6e6f34)) and its fix is as clear as day:

ndcafmaker_mem_x50

The big features (the jumps downward in memory) are correlated in time because the two ND_CAFMaker runs occurred on the same GPVM and started at nearly exactly the same time, so these are kernel memory reallocations.
There might be a very small residual memory leak (you can see the blue line still climbs very gradually) but it could also just be memory fragmentation---RSS is not a perfect gauge of memory leaks.  I'd have to study further with Valgrind's memcheck or similar, and I don't think it's worth the effort at this point.  The obvious, bad memory leak is now obviously plugged.

So, it looks like this PR has done what it's expected to do.  I think it's ready to merge.
</details>

alexbooth92 (approved):
This is _awesome_ @chenel . The prod teams (@jacoblarkin @mjkramer) should definitely be made aware of this, we should be able to up the parallelisation of our CAFMaker jobs!

PR #174, opened 2026-07-16
Eliminate all build warnings, and turn warnings -> errors going forward

chenel:
This PR eliminates all outstanding warnings in the ND_CAFMaker.  Warnings are sometimes harmless/cosmetic, but sometimes surface real issues.  Either way we don't need them in the code.  I've fixed all the ones that were there.

One of them found what I think is a real logic error that might change behavior in the ND-LAr--TMS track matching (20c0069f): what was there was definitely wrong (variables corresponding to some dot products were being set and were never used, instead of updating the variables with the same name that actually were used).  That needs checking by expert eyes (@QWeyrich).

I also removed the now-defunct `NDLArTMSMatchFiller`, since it was fully commented out anyway and is completely superseded by the `NDLArTMSUniqueMatchFiller`.

Since there are now no warnings, I also set the compiler flag that turns warnings into errors, so that we don't get any more errors in future.  (CI now requires an error-free build.)


alexbooth92 (approved):
Lots of discussion on this on slack with @QWeyrich. It is likely that the behaviour of the track matching will change, but that the algorithm's parameters require an update anyway.

